PR #20, opened 2023-10-10
Add pyiron-data as additional dependency

jan-janssen:


liamhuber:
Indeed, everything runs fine with `pyiron-data` explicitly included.

PR #33, opened 2023-10-16
Joerg's lammps nodes

liamhuber:
Just directly pulled over from [the contrib PR](https://github.com/pyiron/pyiron_contrib/pull/856) plus updating the workflow import paths.


liamhuber:
Notebooks test is failing because I disabled registering new packages in order to get macros working on executors. Getting registration working again is my second priority (after a pull method), but I'll hotfix this branch later tonight to give the lammps nodes the same special favour that atomistics and standard nodes have so they're available.

pyiron_workflow/node_library/dev_tools.py
```diff
@@ -0,0 +1,87 @@
...
+    ):
+        self.value = value
+        self.type = dat_type
+        self.label = label
+        self.store = store
+        self.generic = generic
+        self.doc = doc
+
+
+class FileObject:
```
samwaseda, line 25: This is nearly the same object as the [`FileObject`](https://github.com/pyiron/pyiron_workflow/blob/bb79722dcc0d0263f5514fa52c0cd767e501bcf4/pyiron_workflow/files.py#L75) that we already have in [`files.py`](https://github.com/pyiron/pyiron_workflow/blob/main/pyiron_workflow/files.py)

liamhuber:
@JNmpi, just FYI: since you added the demo notebook for me to test against, I plan to stack a patching PR here every time a new version is released so that this PR stays up-to-date. Where possible/where I notice it I'll replace some of the tools with stuff that is in the main infrastructure (things that are sometimes directly inspired by the dev tools here), e.g. the registration. As pointed out by @samwaseda, bits of this duplicate tools that were already in the main part of the code, so I'm just lazily postponing a more time-intensive full merge here -- it's my intent to do one, but I don't want to promise that before we get v1.0.0 with storage.

liamhuber:
I'm also running into some surprising (to me) errors:

Starting with

```python

wf = Workflow("tmp")
wf.p1 = Workflow.create.standard.Add(1)
```

Running linearly works fine, but once I introduce multiple workers,

```python
wf.iter(p1__other=[1,2,3,4], max_workers=2, executor=True)
```

Gives `TypeError: Workflow.__init__() got an unexpected keyword argument 'p1__other'`; my guess is because the re-instantiated workflow doesn't have a `p1` node at all, although I'm not certain yet.

And

```python
wf.p1.iter(other=[1,2,3,4], max_workers=2, executor=True)
```

Gives

```
ReadinessError: Add received a run command but is not ready. The node should be neither running nor failed, and all input values should conform to type hints.
Add readiness: False
STATE:
running: False
failed: False
INPUTS:
obj ready: False
other ready: True
```

Almost certainly because the first argument value is not getting carried along to the new copies. `wf.p1.iter(obj=[1], other=[1,2,3,4], max_workers=2, executor=True)` fails, but with the file error that has to do with trying to merge storage in, which means the input is passing just fine and this confirms my suspicion.

liamhuber:
This content has been migrated to https://github.com/pyiron/node_library, where it currently depends on `pyiron_workflow-0.8.0`. I want to leave this branch around, but I'm going to close the PR.

PR #37, opened 2023-10-18
No io by reference for macros

liamhuber:
Macros are now walled gardens that have their own IO objects instead of just passing children by reference.

I want to make a final test for ensuring that macro connections get properly recreated after rebuilding the macro IO, but it's working in the demo notebook so this should just be a formality. I expect these changes to make it very easy to extend macros so they run on an executor.

TODO:
- [x] Test macro connections are reconstructed when io is rebuild

### DIFF
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -666,8 +666,8 @@
     {
      "data": {
       "text/plain": [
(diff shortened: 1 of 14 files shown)


liamhuber:
Ah, something's missing a getstate -- it's default in 3.11 but I'll need to track it down here for 3.10.

PR #47, opened 2023-10-23
Add a `pull()` method

### LINKED ISSUE #10: Accessing information without full workflow execution
From @JNmpi in pyiron/pyiron_contrib#855 

> #### Provide ways to show dynamic input options
> ```python
> wf = Workflow('simple_lammps_calculation')
> 
> structure = wf.create.atomistics.Bulk(cubic=True, name="Cu", label='structure')
> 
> lammps_job = wf.create.lammps.LammpsStaticNode(label='lammps_job')
> lammps_job.inputs.structure = structure 
> 
> lammps_job.inputs.potential.list   # show all potentials compatible to structure
> lammps_job.list_potentials         # alternative syntax (list_potentials would be a node in a workflow that is not called when running the workflow but only when calling it explicitly)
> ```
> 
>  ```
> potential starts
> ```
> 
> **Issue:** Having nodes that are not part of the main workflow but that can be used by the user to extract/provide information when explicitly called would be very helpful.

Ah, I checked out your branch (#856) and I see what you're saying. You created exactly such a node (`ListPotentialsNode`) and (after I uncomment a line) it exists inside your `LammpsStaticNode` macro.

The problem, of course, is that if I simply run the node -- `lammps_job.list_potentials()` -- it doesn't yet have any input, so the result is an error. If I first run the workflow, it will crash and give an ugly error, _but_, provided the "information node" runs in the macro before the macro crashes (it does in this case), I can look at its output. I.e. on your branch after uncommenting `wf.list_potentials = wf.create.lammps.ListPotentialsNode(structure=wf.structure)` in `lammps_static_node`, I can do this:

```python
from pyiron_contrib.workflow import Workflow
from pyiron_contrib.workflow.node_library.lammps import nodes as lammps_nodes

Workflow.create.register("lammps", *lammps_nodes)

wf = Workflow('simple_lammps_calculation')

structure = wf.create.atomistics.Bulk(cubic=True, name="Cu", label='structure')

lammps_job = wf.create.lammps.LammpsStaticNode(label='lammps_job')
lammps_job.inputs.structure = structure

wf() 
```

Which crashes as expected since I never specified a potential, but populates the following:

```python
wf.lammps_job.list_potentials.outputs.potentials.value
>>> ['1985--Foiles-S-M--Ni-Cu--LAMMPS--ipr1', ...]
```

Certainly this is more hoops that we want the user to jump through -- they should never need to intentionally cause errors to get what they want -- but at least it is _possible_.

Fundamentally, I think we're facing the tension between delayed and aggressive execution again. What we'd really like is for the structure to propagate through so that the `list_potentials` node just has its input right away. While this isn't possible in a fully delayed framework, I do think implementing `pull` (#836) gives us a sufficiently good solution. With `pull`, we can replace the `wf()` call above with the following:

```python
wf.lammps_job.list_potentials.pull()
>>> {'potentials': ['1985--Foiles-S-M--Ni-Cu--LAMMPS--ipr1', ...]}
```

Thus getting the desired result without ever causing the workflow to crash. IMO this would be a satisfactory solution, so any PR that closes pyiron/pyiron_workflow#12 should also close this.

## An aside on node empowerment

An alternative solution is to directly empower nodes by defining them as classes instead of using the decorators. For instance, in this case I could replace the decorated `lammps_static_node` function with

```python
class LammpsStaticNodeWithMethod(Macro):
    def __init__(
        self,
        label: Optional[str] = None,
        parent=None,
        strict_naming: bool = True,
        **kwargs,
    ):
        super().__init__(
            graph_creator=self._make_graph,
            label=label,
            parent=parent,
            strict_naming=strict_naming,
            inputs_map=None,
            outputs_map=None,
            **kwargs,
        )

    @staticmethod
    def _make_graph(wf):
        ...  # All the content of lammps_static_node

    def list_potentials_method(self, structure=None):
        from pyiron_atomistics.lammps.potential import list_potentials
        potential = list_potentials(
            self.inputs.structure.value if structure is None else structure
        )
        return potential
```

Then there's a number of possibilities, but presuming again that `pull` has been implemented, I could do something like this:

```python
from pyiron_contrib.workflow import Workflow
from pyiron_contrib.workflow.node_library.lammps import nodes as lammps_nodes

Workflow.create.register("lammps", *lammps_nodes)

wf = Workflow('simple_lammps_calculation')

structure = wf.create.atomistics.Bulk(cubic=True, name="Cu", label='structure')

lammps_job = wf.create.lammps.LammpsStaticNode(label='lammps_job')
lammps_job.inputs.structure = structure

lammps_job.list_potentials_method(structure.pull().structure)
>>> ['1985--Foiles-S-M--Ni-Cu--LAMMPS--ipr1', ...]
```

This is spiritually similar to how we extended `Node` with `.draw()` method, giving users an extra tool accessible from the node. In this particular example I think I prefer your approach of an information-carrying-sub-node, but there may be situations where adding a method is superior.

### LINKED ISSUE #12: "pull"-mode / delayed behaviour
@JNmpi sold me a while ago on the merits of using a "pull" paradigm instead of a "push" paradigm, but I couldn't see a way to get the technicalities sorted out. I think I see a path through now and will be working on this topic this week.

Most of the classes look the same, and I think we do still need signals for making cyclic graphs work (and indeed, switches in general), but I think this will greatly streamline things and make "easy things easy" while not making hard things any harder than they already are.

Here's my thoughts so far on the class behaviour:

- `OutputData.pull()`
  - Calls `self.parent.outputs.pull(): Outputs.pull()`
  - `Outputs.pull()` manually sets `OutputData.value` for each child
  - Return `self.value` once the `Outputs.pull()` call resolves
- `Outputs.pull()`
  - Calls `self.node.pull(): Node.pull()`
  - Uses the returned dictionary to update the values of all channels
- `Node.pull()`
  - Calls `self.inputs.pull(): Inputs.pull()`
  - Hashes the resulting dictionary of input data 
  - Compares the hash to the hash for the last call...
    - If there is no such hash, or hashes differ, run with input, returns result of run function
    - Else return existing output `self.outputs.to_dict()`
    - This prevents unnecessary running when input hasn't changed
    - Need a flag like `Node.always_rerun` in case we _want_ to run again on the same input! E.g. random number generation
- `Inputs.pull()`
  - Calls `.pull()` on each `InputData` child
  - Returns children's values as a dictionary for use in the node
- `InputData.pull()`
  - For each connection in `self.connections: list[OutputData]` call `.pull()`
    - The first one that returns something other than `NotData` gets used
    - _Somewhere_ we need to do more clever grouping, so that we directly call `Outputs.pull()` so that groups of output can get updated at once without multiple graph traversals where we discover that the input hasn't changed...
    - This is a change that makes `connections` and _ordered list_ so we will always first try to draw data from the 0th connection, then use the 1st as a fall-back, etc.
  - If `len(self.connections) == 0` return `self.value`
    - If no default exists, and no value was set manually, e.g. by the user of the workflow, the initial `NotData` value for `DataChannel.value` gets returned -- obviously this will break something somewhere, either by raising an error or by keeping some form of the `Channel/IO/Node.ready` attribute that will stop execution.

In this way, we update input when we invoke a node (`my_wf(some_node__some_channel=42)`) as we currently do, but then instead of starting a "push" process from the "upstream-most" nodes (or those specified in `starting_nodes`), we start a "pull" process from the "downstream-most" nodes (or those specified in `result_nodes` or even maybe `result_channels` or similar).

I've got to leave it there for tonight, but will update this tomorrow and as I work on the corresponding PR.

### LINKED ISSUE #18: The run cycle needs refactoring
I went back to start extracting stuff for a pull-mode, and I found a few too many levels of mis-direction. I'd like to revisit `run` and break it into different component hopefully with fewer layers of misdirection.

There is nothing really philosophically new here compared to #12, especially including [this comment](https://github.com/pyiron/pyiron_workflow/issues/12#issuecomment-1753185935) about replacing `OutputData._after_update` with some sort of `InputData.pull`, I just wanted to give an explicit heads up that the API around `Node.run/run_args/on_run/finish_run` is likely to change!

### LINKED ISSUE #35: Simple example even simpler
**Summary**

Currently, the `pyiron_workflow` code is still more complex than other frameworks.

**Detailed Description**

Example from the `README.MD`: 
```
wf = Workflow("my_workflow")
wf.a1 = x_plus_y()
wf.a2 = x_plus_y()
wf.b = x_plus_y(x=wf.a1.outputs.sum, y=wf.a2.outputs.sum)

out = wf(a1__x=0, a1__y=1, a2__x=2, a2__y=3)
out.b__sum
>>> 6
```

I would like to write: 
```
out = x_plus_y(
    x=x_plus_y(x=0, y=1),
    y=x_plus_y(x=2, y=3),
)
out.sum
```

**Further Information, Files, and Links**

As far as I understand this is currently related to enforcing labels for the individual nodes of the workflow, so the suggestion would be to generate labels automatically. I am aware that labels can be helpful for debugging the workflow, but I would like to have a certain level of compatibility to existing python functions.

liamhuber:
Before getting to the meat of it, I updated the API so that all the actual computations using the node get based off the `run` method, which is now made more flexible with some boolean arguments:
- `run(fetch_input_first: bool = True, then_emit: bool = True, force_local_execution: bool = False, check_readiness: bool = True,)` fetches input from connections, raises an error if it's not `ready`, computes (either locally or on an executor), and finally fires the `ran` signal
- `execute()`  to force the computation right here, right now, as-is with `run(fetch_input_first=False, then_emit=False, force_local_execution=True, check_readiness=False)`
- `__call__(*args, **kwargs)` to update input values with (*args and) **kwargs then `run(fetch_input_first=True, then_emit=False)`
- `pull()` to build an independent run cycle, run upstream nodes, then `run(fetch_input_first=True, then_emit=False)`
- ~`update()`~ is just not a thing, we handle it slightly differently and as a flag in `run()`

This closes #18 to my satisfaction.

Next, all the topology stuff was pulled out of `pyiron_workflow.composite.Composite` and into `pyiron_workflow.topology`.

Finally, implement `Node.pull` which does as above. I had to do a little bit of node relabelling to get it to work when the nodes are outside a workflow (since then they can have arbitrary -- even identical -- labels), but we reset everything after the pull is done anyhow.

Closes #35 because the following now works:

```python

@Workflow.wrap_as.single_value_node("sum")
def x_plus_y(x: int = 0, y: int = 0) -> int:
    return x + y

node = x_plus_y(
    x=x_plus_y(0, 1), 
    y=x_plus_y(2, 3)
)
node.pull()
>>> 6
```

Closes #10 because you can now `pull` on just one part of a macro. The caveat here is that since macros are now "walled gardens" with their own IO, pulling on a macro's child will only go as far upstream as the macro input (to which the child input is linked), so if you want fresher input that that you need to `.set_input_values` on the parent macro, or `.pull()` the entire macro, or just the upstream bit you want to freshen up followed by `.inputs.fetch()` on the macro. Example:

```python

@Workflow.wrap_as.single_value_node("sum")
def x_plus_y(x: int = 0, y: int = 0) -> int:
    print("EXECUTING")
    return x + y

@Workflow.wrap_as.macro_node()
def b2_leaves_a1_alone(macro):
    macro.a1 = x_plus_y(0, 0)
    macro.a2 = x_plus_y(0, 1)
    macro.b1 = x_plus_y(macro.a1, macro.a2)
    macro.b2 = x_plus_y(macro.a2, 10)
    
wf = Workflow("demo")
wf.upstream = x_plus_y()
wf.macro = b2_leaves_a1_alone(a2__x=wf.upstream)


# Pulling b1 -- executes a1, a2, b2
print(wf.macro.b1.pull())
>>> EXECUTING
>>> EXECUTING
>>> EXECUTING
>>> 1

# Pulling b2 -- executes a2, a1
print(wf.macro.b2.pull())
>>> EXECUTING
>>> EXECUTING
>>> 11

# Updated inputs get reflected in the pull
wf.macro.set_input_values(a1__x=100, a2__x=-100)
print(wf.macro.b2.pull())
>>> EXECUTING
>>> EXECUTING
>>> -89

# Connections are restored after a pull
# Running wf -- executes upstream, macro (is silent), a1, a2, b1, b2
print(wf())
>>> EXECUTING
>>> EXECUTING
>>> EXECUTING
>>> EXECUTING
>>> EXECUTING
>>> {'macro__b1__sum': 101, 'macro__b2__sum': 11}
```

And of course it closes #12, the original issue on the topic, because `pull()` is now implemented. The technical realization is actually quite close to the 21 September comment there as well.

This isn't as thoroughly tested as I'd like, but these examples are included in the integration test -- so at least the successful cases are behaving as expected. I will have time to write additional tests this week, but not today ahead of the env deadline for the ADIS workshop.

### DIFF
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -119,8 +119,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# pm_node.run()\n",
-    "pm_node.update()"
+    "# pm_node.run()"
    ]
   },
   {
@@ -128,9 +127,7 @@
    "id": "48b0db5a-548e-4195-8361-76763ddf0474",
    "metadata": {},
    "source": [
-    "Using the softer `update()` call checks to make sure the input is `ready` before moving on to `run()`, avoiding this error. In this case, `update()` sees we have no input an aborts by returning `None`.\n",
-    "\n",
-    "(Note: If you _do_ swap `update()` to `run()` in this cell, not only will you get the expected error, but `pm_node` will also set its `failed` attribute to `True` -- this will prevent it from being `ready` again until you manually reset `pm_node.failed = False`.)"
+    "Not only will you get the expected error, but `pm_node` will also set its `failed` attribute to `True` -- this will prevent it from being `ready` again until you manually reset `pm_node.failed = False`."
    ]
   },
   {
@@ -176,7 +173,7 @@
    "id": "c54a691e-a075-4d41-bc0f-3a990857a27a",
    "metadata": {},
    "source": [
-    "Alternatively, the `run()` command (and `update()` when it proceeds to execution) just return the function's return value:"
+    "Alternatively, the `run()` command just return the function's return value:"
    ]
   },
   {
@@ -241,7 +238,7 @@
    "id": "58ed9b25-6dde-488d-9582-d49d405793c6",
    "metadata": {},
    "source": [
-    "This node also exploits type hinting! `run()` will always force the execution, but `update()` will not only check if the data is there, but also if it is the right type:"
+    "This node also exploits type hinting! `run()` will check that input values conform to type hints before computing anything. Failing at this stage won't actually cause the node to have a `failed` status, so you can just re-run it once the input is fixed."
    ]
   },
   {
@@ -249,46 +246,27 @@
    "execution_count": 10,
    "id": "ac0fe993-6c82-48c8-a780-cbd0c97fc386",
    "metadata": {},
-   "outputs": [],
-   "source": [
-    "adder_node.inputs.x = \"not an integer\"\n",
-    "adder_node.inputs.x.type_hint, type(adder_node.inputs.x.value)\n",
-    "adder_node.update()\n",
-    "# No error because the update doesn't trigger a run since the type hint is not satisfied"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "id": "2737de39-6e75-44e1-b751-6315afe5c676",
-   "metadata": {},
-   "source": [
-    "Since the execution never happened, the output is unchanged"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": 11,
-   "id": "bcbd17f1-a3e4-44f0-bde1-cbddc51c5d73",
-   "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "1"
+       "(int, str)"
       ]
      },
-     "execution_count": 11,
+     "execution_count": 10,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
-    "adder_node.outputs.sum_.value"
+    "adder_node.inputs.x = \"not an integer\"\n",
+    "adder_node.inputs.x.type_hint, type(adder_node.inputs.x.value)\n",
+    "# adder_node.run()"
    ]
   },
   {
    "cell_type": "code",
-   "execution_count": 12,
+   "execution_count": 11,
    "id": "15742a49-4c23-4d4a-84d9-9bf19677544c",
    "metadata": {},
    "outputs": [
@@ -298,14 +276,14 @@
        "3"
       ]
      },
-     "execution_count": 12,
+     "execution_count": 11,
      "metadata": {},
      "output_type": "execute_result"
     }
    ],
    "source": [
     "adder_node.inputs.x = 2\n",
-    "adder_node.update()"
+    "adder_node.run()"
    ]
   },
   {
@@ -318,7 +296,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 13,
+   "execution_count": 12,
    "id": "0c8f09a7-67c4-4c6c-a021-e3fea1a16576",
    "metadata": {},
    "outputs": [
@@ -328,7 +306,7 @@
        "30"
       ]
      },
-     "execution_count": 13,
+     "execution_count": 12,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -348,7 +326,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 14,
+   "execution_count": 13,
    "id": "69b59737-9e09-4b4b-a0e2-76a09de02c08",
    "metadata": {},
    "outputs": [
@@ -358,7 +336,7 @@
        "31"
       ]
      },
-     "execution_count": 14,
+     "execution_count": 13,
      "metadata": {},
      "output_type": "execute_result"
     }
@@ -391,7 +369,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": 14,
    "id": "61b43a9b-8dad-48b7-9194-2045e465793b",
    "metadata": {},
    "outputs": [],
@@ -401,7 +379,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 16,
+   "execution_count": 15,
    "id": "647360a9-c971-4272-995c-aa01e5f5bb83",
    "metadata": {},
    "outputs": [
@@ -438,7 +416,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 17,
+   "execution_count": 16,
    "id": "b8c845b7-7088-43d7-b106-7a6ba1c571ec",
    "metadata": {},
    "outputs": [
@@ -482,7 +460,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 18,
+   "execution_count": 17,
    "id": "2e418abf-7059-4e1e-9b9f-b3dc0a4b5e35",
    "metadata": {
     "tags": []
@@ -532,7 +510,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 19,
+   "execution_count": 18,
    "id": "59c29856-c77e-48a1-9f17-15d4c58be588",
    "metadata": {},
    "outputs": [
@@ -568,7 +546,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 20,
+   "execution_count": 19,
    "id": "1a4e9693-0980-4435-aecc-3331d8b608dd",
    "metadata": {},
    "outputs": [],
@@ -580,7 +558,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 21,
+   "execution_count": 20,
    "id": "7c4d314b-33bb-4a67-bfb9-ed77fba3949c",
    "metadata": {},
    "outputs": [
@@ -619,7 +597,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 22,
+   "execution_count": 21,
    "id": "61ae572f-197b-4a60-8d3e-e19c1b9cc6e2",
    "metadata": {},
    "outputs": [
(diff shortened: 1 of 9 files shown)


liamhuber:
`Push-main-Pull-all / pyiron / unit-tests (ubuntu-latest, 3.11, linux-64-py-3-11, /usr/share/...` and `Push-main-Pull-all / pyiron / build-notebooks (pull_request)` are both hanging. They each have a message like `Received 959348001 of 963542305 (99.6%), 0.1 MBs/sec` that repeats every tick, so I think it's just a communication hiccup. I'll restart these two.

liamhuber:
Yeah, no problems on the rerun

JNmpi:
Thanks @liamhuber for the implementation. The new functionality and options are great and provide much for flexibility. I tried to merge my latest branch with hotfix0.0.2 (containing the changes you did on my branch) with main. When running my notebook in this updated branch I get the following error messages:

### MD over list of temperatures


```python
%%time
for T in [300, 600, 900]:
    wf = Workflow('Lammps')
    structure = wf.create.atomistics.Bulk('Al', cubic=True)
    repeat = wf.create.lammps.Repeat(structure = structure, repeat_scalar=3)
    
    lammps = wf.create.atomistic_codes.Lammps(structure=repeat, label='lammps')
    lammps.calc_select.md(temperature=T, n_ionic_steps=10_000)
    
    wf.run()
    
    plt.plot(wf.lammps.outputs.generic.value.energy_pot[2:], label=f'T={T}K')
plt.legend();    
```

    /Users/joerg/python_projects/git_libs/pyiron_workflow/pyiron_workflow/channels.py:158: UserWarning: The channel calculator was not connected to calculator, andthus could not disconnect from it.
      warn(
    /Users/joerg/python_projects/git_libs/pyiron_workflow/pyiron_workflow/channels.py:158: UserWarning: The channel ran was not connected to run, andthus could not disconnect from it.
      warn(
    /Users/joerg/python_projects/git_libs/pyiron_workflow/pyiron_workflow/channels.py:158: UserWarning: The channel structure was not connected to structure, andthus could not disconnect from it.
      warn(



    ---------------------------------------------------------------------------

    AttributeError                            Traceback (most recent call last)

    File <timed exec>:7


    File ~/python_projects/git_libs/pyiron_workflow/pyiron_workflow/node_library/dev_tools.py:63, in Replacer.__call__(self, other, **kwargs)
         61     # print ('replacer called', self.node.label, other, kwargs) #, other.info, dir(other))
         62     self.node = self.parent[self.node_label]
    ---> 63     return self.node.update_input(**kwargs)
         65 else:
         66     import warnings


    File ~/python_projects/git_libs/pyiron_workflow/pyiron_workflow/function.py:610, in SingleValue.__getattr__(self, item)
        609 def __getattr__(self, item):
    --> 610     return getattr(self.single_value, item)


    AttributeError: type object 'NotData' has no attribute 'update_input'



and 

```python
import numpy as np

energy_pot = []
strain_lst = np.linspace(0.86, 2, 51)
for strain in strain_lst:
    wf = energy_at_volume(element='Al', strain=strain)
    out = wf.run()

    energy_pot.append(out['lammps__generic'].energy_pot.iloc[0])  

plt.plot(strain_lst, energy_pot);
```


    ---------------------------------------------------------------------------

    ValueError                                Traceback (most recent call last)

    Cell In[7], line 7
          5 for strain in strain_lst:
          6     wf = energy_at_volume(element='Al', strain=strain)
    ----> 7     out = wf.run()
          9     energy_pot.append(out['lammps__generic'].energy_pot.iloc[0])  
         11 plt.plot(strain_lst, energy_pot);


    File ~/python_projects/git_libs/pyiron_workflow/pyiron_workflow/node.py:281, in Node.run(self, first_fetch_input, then_emit_output_signals, force_local_execution, check_readiness)
        277 if check_readiness and not self.ready:
        278     input_readiness = "\n".join(
        279         [f"{k} ready: {v.ready}" for k, v in self.inputs.items()]
        280     )
    --> 281     raise ValueError(
        282         f"{self.label} received a run command but is not ready. The node "
        283         f"should be neither running nor failed, and all input values should"
        284         f" conform to type hints:\n"
        285         f"running: {self.running}\n"
        286         f"failed: {self.failed}\n" + input_readiness
        287     )
        288 return self._run(
        289     finished_callback=self._finish_run_and_emit_ran
        290     if then_emit_output_signals
        291     else self._finish_run,
        292     force_local_execution=force_local_execution,
        293 )


    ValueError: energy_at_volume received a run command but is not ready. The node should be neither running nor failed, and all input values should conform to type hints:
    running: False
    failed: False
    element ready: True
    bulk__crystalstructure ready: True
    bulk__a ready: True
    bulk__c ready: True
    bulk__covera ready: True
    bulk__u ready: True
    bulk__orthorhombic ready: True
    bulk__cubic ready: True
    repeat__repeat_scalar ready: True
    strain ready: True
    lammps__potential ready: True
    lammps__potential__index ready: True
    lammps__init_lammps__working_directory ready: True
    lammps__shell__command ready: False
    lammps__shell__environment ready: False
    lammps__shell__arguments ready: False

Do you have an idea what is going on? I guess it is related to changing the default values in the run environment and choosing them properly would fix the error. Probably you see much faster what is going on than me.

liamhuber:
@JNmpi, ~I'm afraid I can't reproduce those errors~ yes, I can reproduce these after merging main (v0.0.3) into the hotfix branch as well.

```
...
File ~/python_projects/git_libs/pyiron_workflow/pyiron_workflow/node_library/dev_tools.py:63, in Replacer.__call__(self, other, **kwargs)
...
---> 63     return self.node.update_input(**kwargs)
...
AttributeError: type object 'NotData' has no attribute 'update_input'
```

This one is pretty easy; I renamed `update_input` to `set_input_values`, but `Replacer` didn't exist on the branch I did the name change on so it never got updated. I'll patch that now.

```
...
----> 7     out = wf.run()
...
ValueError: energy_at_volume received a run command but is not ready. The node should be neither running nor failed, and all input values should conform to type hints:
...
lammps__shell__command ready: False
lammps__shell__environment ready: False
lammps__shell__arguments ready: False
```

This is coming up because the `shell` sub-node input is either `NotData` or not matching its type hint, both of which are disallowed prior to running. This can be overridden with `.run(check_readiness=False)`, but it is better and probably not too hard to just fix the root cause of the problem.

For `environment` and `arguments` I think it's really simple -- these are optional, and if they're `None` they're correctly set to empty mutable defaults. Adding `dict` --> `Optional[dict]` (and similarly for `list`) should do the trick. I'll patch it now.

It took a minute longer for `command`, but it's actually basically the same issue -- it's hinted as `str` but the `lammps` macro sets it to an instance of `ExecutablePathResolver`. I think just adding `.path()` to this initial setting will fix it. I'll patch that too.

I stacked the patches onto the hotfix here: #49. 

That last error would have been much easier to see if we type-checked value assignments. The rationale for _not_ doing this is that type checking is a little expensive, and I don't want the automated data propagation doing it all the time; thus only new channel connections are type checked. 

I was thinking this week that a better way is probably to do the type checking by default -- so users (and usually devs) will get immediate feedback when they try to assign something illegal -- but have an "escape hatch" we can use in `InputData.fetch` so that the automated processes skip this step.

I'll make an issue to this effect.

PR #53, opened 2023-10-26
Fix node registration

### LINKED ISSUE #51: (Re)allow registering new node packages
For good measure I should make an issue for this. Node package registration, i.e. `wf.create.register("my_nodes", *my_nodes)` got turned off because the implementation was not wise and didn't play well with executors (like, at all). There's nothing wrong with the idea, of course. Joerg also has a nice modification for registering modules over on #33. I also promised somewhere to create an alias so you can just `wf.register`. It shouldn't be hard, it just needs to be done. Now that `Node.pull` is working it is my top priority.

liamhuber:
Adds the node registration method back, although at the moment it's just called in `Creator.__init__` so you don't see it.

TODO:
- [x] Tests
- [x] Think about what API you want to support -- should we also handle lists of nodes, or modules containing a `nodes` attribute (as long as it has the right type), or URLs? (EDIT: Right now I'm still only accepting string import paths to modules that have a `nodes: list[Node]` property, but the `Creator.register` method has good encapsulation of the logic for checking whether (a) newly proposed packages are just a re-registration of an existing package, and (b) whether the proposed package will actually be usable for generating nodes, so extending this in the future to other node package identifiers should be easy. For now I want to prioritize getting _any_ fix in so we at least have some way to register things!)
- [x] Stop registering `atomistics` by default and update the demo and docs as needed
- [x] More gracefully handle re-registering things -- if it's the same nodes being set to the same domain name, just let it pass silently! This may be helpful when...
- [x] We encourage somewhere in the docs that macros register the necessary packages _inside_ the macro definition whenever the macro uses `.create...` (i.e. unless the macro is part of the same .py file that all its children are coming from, in which case they should be available in-scope without `.create`

Outside of scope:
- *Actually* supporting all manner of registration types (like a URL) -- I just want infrastructure that will be flexible enough to handle that in the future, not to get it all solved now!

Closes #51

### DIFF
--- a/notebooks/workflow_example.ipynb
+++ b/notebooks/workflow_example.ipynb
@@ -644,8 +644,8 @@
     {
      "data": {
       "text/plain": [
-       "array([0.91077351, 0.33860412, 0.59806048, 0.66528464, 0.80125293,\n",
-       "       0.31981677, 0.54395521, 0.4926537 , 0.52626431, 0.7848854 ])"
+       "array([0.49455794, 0.6789772 , 0.48470916, 0.43574953, 0.18030331,\n",
+       "       0.6059215 , 0.65871187, 0.42205006, 0.65062977, 0.5390317 ])"
       ]
      },
      "execution_count": 22,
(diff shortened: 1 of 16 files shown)


liamhuber:
The 3.9 tests break:

```
...
TypeError: unsupported operand type(s) for |: 'type' and 'type'

...
ValueError: The package identifier is pyiron_workflow.node_library.standard is not valid. Please ensure it is an importable module with a list of Node objects stored in the variable `nodes`.
```

At first I was confused why importing from `pyiron_workflow.node_library.standard` would not work, but scrolling up the trace it's actually breaking when trying to process the type hint. This makes sense since 3.9 doesn't handle `|` but requires `typing.Union`, hence why we brute-force skip almost all the tests until the version is >= 3.10.

This failure happens in:
```
ERROR: executors.test_cloudprocesspool (unittest.loader._FailedTest)
ERROR: test_macro (unittest.loader._FailedTest)
ERROR: test_util (unittest.loader._FailedTest)
ERROR: test_workflow (unittest.loader._FailedTest)
```

In each case starting with

```
ImportError: Failed to import test module: [test module name]
```

So I guess I've somehow added a "bad" type hint somewhere before we get to the `skipif` version guard.

liamhuber:
The problem is when `Creator` tries to register the `standard` node package, which makes enough sense since we parse the type hints there and definitely use the new style.

What doesn't yet make sense is why there's a creator existing in, e.g., the `tests/unit/executors/test_cloudprocesspool`? And since the `standard` registration happens in `__init__` is there really one of these being initialized, or is this some strange side-effect of the `Creator` being a `Singleton`?

liamhuber:
Ah, ok, I think maybe because right in `pyiron_workflow.__init__` we import `Workflow`, which inherits from `Composite`, which instantiates a `Creator` as a _class attribute_.

We no longer get away with this because now we `typing.get_type_hints` as part of the decorator creating new node classes.

Alright, so all perfectly sensible now how to fix it.

liamhuber:
```
ERROR: test_timeout (executors.test_cloudprocesspool.TestCloudpickleProcessPoolExecutor.test_timeout)
  File "/Users/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/executors/test_cloudprocesspool.py", line 158, in test_timeout
    fs.result(timeout=30),
    ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/work/pyiron_workflow/pyiron_workflow/pyiron_workflow/executors/cloudpickleprocesspool.py", line 10, in result
    result = super().result(timeout=timeout)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/miniconda3/envs/my-env/lib/python3.11/concurrent/futures/_base.py", line 458, in result
    raise TimeoutError()
TimeoutError

```

Here we wait 30s and expect a result. It passed for other versions/OSs, so I think things were just super slow for some reason. I'm going to be a little more generous with the waiting time and try again.

liamhuber:
Notebook failure is expected since I stopped registering atomistics by default and broke the scatter plot node over to its own package.

liamhuber:
> Codacy Coverage Report — -1.15% coverage variation (-1% target)

Coverage is just angry because `pyiron_workflow.node_library.atomistics` is no longer getting caught in the unit tests, but that's kind of the point; the unit tests are meant to test the infrastructure not the nodes we build on top of it. When the nodes get spun off to their own repos they can have their own tests.

PR #59, opened 2023-10-31
Pull recursively

### LINKED ISSUE #54: pull mode should be allowed to (always?) pull parent macro input first
Thanks @liamhuber for implementing the pull mode. The examples you provided work nicely but when I try to run it for the lammps example it fails. The code for the calling cell is:

          wf = Workflow('Lammps')
          structure = wf.create.atomistics.Bulk('Al', cubic=True)
          repeat = wf.create.lammps.Repeat(structure = structure, repeat_scalar=3)
          
          lammps = wf.create.atomistic_codes.Lammps(structure=repeat, label='lammps')
          repeat.pull()  # works
          # lammps.structure.pull() # does not work
          # lammps.list_pots.pull() # does not work
          
          # lammps() # works

The lammps node in atomistic_codes is:

        @macro_node()
        def lammps(wf: Macro) -> None:
            from pyiron_contrib.tinybase.shell import ExecutablePathResolver
        
            structure = wf.create.lammps.Structure()
        
            potential = wf.create.lammps.Potential(structure=structure)
        
            wf.list_pots = wf.create.lammps.ListPotentials(structure=structure)

The rest of the node is identical to the one in the atomistic_codes library. Do you have any idea what is going on and why the above code fails. 

When writing the example I realized that it would be nice to add optional functions to a macro which are only called via an explicit `pull()`, but not if we run the workflow. A possible syntax would be:

      wf.add_function(list_potential=wf.create.lammps.ListPotentials(structure=wf.structure))

liamhuber:
Allow `pull` to recursively break out of its own context and cause parent nodes to also pull _their_ input data trees. 

`pull`, `execute`, and `__call__` are all just shortcuts for `run` with a convenient set of flags pre-set. The `run` command has a bunch of boolean flags to control the type of execution. To summarize these I'll just copy-paste the new docstring section:

```
Args:
    run_data_tree (bool): Whether to first run all upstream nodes in the data
        graph. (Default is False.)
    run_parent_trees_too (bool): Whether to recursively run the data tree in
        parent nodes (if any). (Default is False.)
    fetch_input (bool): Whether to first update inputs with the
        highest-priority connections holding data. (Default is True.)
    check_readiness (bool): Whether to raise an exception if the node is not
        `ready` to run after fetching new input. (Default is True.)
    force_local_execution (bool): Whether to ignore any executor settings and
        force the computation to run locally. (Default is False.)
    emit_ran_signal (bool): Whether to fire off all the output `ran` signal
        afterwards. (Default is True.)
    **kwargs: Keyword arguments matching input channel labels; used to update
        the input channel values before running anything.
```

Here `__call__` is now the most aggressive thing imaginable, which invokes runs in both its own upstream data dependencies and (recursively!) parent node data dependencies. (It doesn't trigger downstream runs by emitting its `ran` signal though). 

Note that since `Workflow` instances are a parent-most object, their `run` method cannot run upstream data trees or push downstream execution signals, so in this case `pull` and `run` operate the same (well, run has a couple of other flags you can change, but I mean with default values).

Closes #54.

TODO:
- Extend tests
  - [x] Pulling from inside a macro with and without the recursive call
  - [x] Can we move tests from integration to unit? Yes
  - [x] Test recovery (connections and labels) after failure at various points
  - [x] Figure out how recursive input pulling needs to interact with executors and make sure it is. EDIT: For now, I opted to fail hard if you try to pull when any node in the upstream tree uses an executor. The presence of the executors indicates to me that you might be sitting around waiting for the result longer than expected; IMO it's healthier to push users towards actually putting things in a workflow and then invoking the workflow when they get to the point that they want to be using executors.
- [x] Update example notebook
- [x] Update docstring examples


liamhuber:
Coveralls failure:

```
...
HTTP error:
---
Error: Bad Gateway (502)
...
```

I need to make more commits here anyway, so let's delay investigating this in case it disappears on its own.

PR #70, opened 2023-11-06
Spec and test

### LINKED ISSUE #61: Add clearer documentation
Implementing `Macro` as a walled garden and `Node.pull` has helped clarify some of the specifications and class interactions in my mind -- in particular, I think I can now say more clear things about the run cycle, the concept of "upstream", and the concept of "parent". This includes the ideas about intended use and `Workflow` being "parent-most" and living and flexible while `Macro` is more crystallized, efficient, and suitable for composition (since it does not have the requirement to be parent-most).  Some of this needs to be translated into class and module docs, but I think a very simple version should be added to the README too.

Additionally, the example notebook needs to be broken down into something like a "quickstart" for people who just want to get running, and a deeper dive for people who want to quickly learn the ins and outs of the codebase.

### LINKED ISSUE #62: Rationalize test suite
Currently there are a number of important abstract classes (namely `Node`, `Composite`, and `Channel`) which hold important shared behaviour for children. Most of this shared behaviour is covered by the tests, but right now it's often tested in an arbitrary child (`Macro` and `Workflow` are the worst for this). 

I'm not sure this was the absolute worst choice to make when the architecture was changing so rapidly, but as things slowly settle down it's a choice that makes the tests very difficult to read. If a new dev comes along (or just future me...) and wants to know where they should put tests for a new feature on one of these base classes, or wants to find the tests for some piece of shared behaviour to gain additional insight into intent, it's very hard to know where to look.

I lean towards having `test_module.py` for each `pyiron_workflow.module`, including for abstract classes. For abstract classes, inside the test environment some trivial child should be created and used to test the shared (held-by-abstract-parent) behaviour. This improve readability and utility of the test suite, even if the coverage is basically the same.

liamhuber:
Update the documentation and testing.

- [x] `Channel`
- [x] `IO`
- [x] `Node`
- [x] `Function`
- [x] `Composite`
- [x] `Macro`
- [x] `Workflow`
- [x] `__init__`
- [x] Make a `quickstart.ipynb`
- [x] Update the current example to re-phrase it as a deep-dive (or maybe just remove it?)
- [x] Update the github README

Closes #61 
Closes #62


liamhuber:
The notebooks failure is actually the same problem that is causing the `pyiron_atomistics` daily tests to fail:

```
File /usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/emmet/core/tasks.py:10
      8 from typing import Any, Dict, List, Optional, Tuple, Type, TypeVar, Union
---> 10 from custodian.vasp.jobs import VaspJob
     11 import numpy as np

ModuleNotFoundError: No module named 'custodian'
```

My knee-jerk suspicion is that materials project has again inserted a direct dependency into their code without including that dependency directly and the API has gotten out of sync with what conda is actually installing, although just because that happened before doesn't mean it's for sure what's happening now.

PR #77, opened 2023-11-15
Use the pympipool executor

### LINKED ISSUE #75: Support `pympipool`
Now that preliminary tests show the `pympipool` executor is working fine for `pyiron_workflow` nodes, I want to expand the executor support from a simple `bool` to something that accepts executor instances/parses instructions to build an executor -- with explicit support and encouragement to use the `pympipool` executor.

To keep things super lightweight I'm planning to have this be an optional dependency and only actually test it in one of the notebooks, but it's time to add the infrastructure.

liamhuber:
Closes #75 

Replaces `Node.executor = True/False` with providing actual executor instances (while keeping infrastructure support for instead instructing _how_ to build an executor in the future, for nesting/reconnection/submitters).

The only big catch right now is that the tests hang on my local machine with the pympipool executor (while they run fine for the built-in and much simpler/weaker `CloudpickleProcessPoolExecutor`). This happens whenever a macro/workflow with children is the one getting an executor -- all the tests where function nodes have an executor work totally fine with both executors. Since this might be related to [pympipool tests hanging on mac](https://github.com/pyiron/pympipool/issues/191), I'm just going to naively push this PR and see if the tests pass OK on linux and windows. Additionally, the hanging tests work totally fine when run in a jupyter notebook -- so I'm optimistic this is something that can be overcome, I just need more data to figure out how.

TODO:
- [ ] Update docs
- [ ] Update demo notebook (at least the quickstart still uses `executor = True`
- [ ] Go look at `pympipool` for how the flux and slurm guys get tested and try to do something similar here


liamhuber:
```
File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pympipool/backend/serial.py", line 30, in main
    input_dict = interface_receive(socket=socket)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.10/site-packages/pympipool/shared/communication.py", line 147, in interface_receive
    return cloudpickle.loads(socket.recv())
ModuleNotFoundError: No module named 'test_macro'
```

This is the cause of failure for a couple of these. Might be those awkward expectations `pympipool` has on the PYTHONPATH that I mentioned somewhere else (the linked issue maybe?)

liamhuber:
Ah, perhaps it runs in the notebook because it recognizes that things are being defined in `__main__` and so knows to use by-value serialization, but when I'm running the tests they're in a .py file and so thinks that by-reference serialization will it doesn't realize that somehow this py file is not actually in the python path.

liamhuber:
Ok, failures are from referencing objects defined in the same file but different scopes, such that the executor thinks it's something that should be available by reference but really it still needs to be pickled by value. Cf. [this issue over on `pympipool`](https://github.com/pyiron/pympipool/issues/235) for a minimal(ish) example.

I can get around this by modifying the existing tests so they never have this same-file-different-scope problem, but I don't want some user/node dev to run into this problem later so let's see if it can be intelligently fixed upstream before giving up.

liamhuber:
Failure was _already_ resolved by pympipool! I was just using an outdated version

liamhuber:
Nope, tests still hang and when I cancel them I get 

```
....Traceback (most recent call last):
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/backend/serial.py", line 64, in <module>
    main(argument_lst=sys.argv)
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/backend/serial.py", line 30, in main
    input_dict = interface_receive(socket=socket)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/shared/communication.py", line 147, in interface_receive
    return cloudpickle.loads(socket.recv())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'test_macro'
```

The pympipool also causes the test suite to hang on my local machine, although 100% of the tests register as passing -- it just doesn't shut down. It's not clear to me whether these two things are connected or not.

liamhuber:
Pycharm no longer hangs, but all the CI tests here still do.

liamhuber:
On cancellation, each of the unit tests fails with

```
File "/Users/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/backend/serial.py", line 64, in <module>
    main(argument_lst=sys.argv)
  File "/Users/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/backend/serial.py", line 30, in main
    input_dict = interface_receive(socket=socket)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/runner/miniconda3/envs/my-env/lib/python3.11/site-packages/pympipool/shared/communication.py", line 147, in interface_receive
    return cloudpickle.loads(socket.recv())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'test_macro'
```

While the coverage fails similarly but with `unit` instead of `test_macro` (this makes enough sense, since the unit tests are being invoked from tests/unit, while the coverage gets invoked directly from tests).

The correct version of pympipool (0.7.9) is getting installed according to the logs, and the versions of mpi4py (3.1.5), cloudpickle (3.0.0), and pyzmq (25.1.1), tqdm (4.66.1) also all match the current pympipool env files and my local environment.

One thing I find particularly interesting is that the unit test failure occurs in `test_macro`; alphabetically `test_function` creates an executor first, _but_ this test raises an exception before we ever use that executor to call `submit`, so the failure is occurring at the first instance of the executor being used.

liamhuber:
That did nothing, I'm getting the exact same `ModuleNotFoundError`s. Let's take one more crack at it from the env side.

liamhuber:
The path is as expected. Maybe the path needs to be added in another process too? I am now trying just manually using a `pympipool` test (mostly), where we first `cd` and then discover tests in `.`.

liamhuber:
Ok, it looks like the `cd` then discover `.` worked -- it ran into errors discovering the static package, but it finished cleanly and the only errors had nothing to do with the executor:

```python
ERROR: test_creator_access_and_registration (test_composite.TestComposite.test_creator_access_and_registration)
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 205, in _verify_identifier
    module = import_module(package_identifier)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/test/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'static'

  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/test_composite.py", line 71, in test_creator_access_and_registration
    self.comp.register("demo", "static.demo_nodes")
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/composite.py", line 517, in register
    cls.create.register(domain=domain, package_identifier=package_identifier)
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 163, in register
    self._verify_identifier(package_identifier)
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 212, in _verify_identifier
    raise ValueError(
ValueError: The package identifier is static.demo_nodes is not valid. Please ensure it is an importable module with a list of Node objects stored in the variable `nodes`.

ERROR: test_registration (test_interfaces.TestCreator.test_registration)
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 205, in _verify_identifier
    module = import_module(package_identifier)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/miniconda3/envs/test/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1126, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1140, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'static'

  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/test_interfaces.py", line 27, in test_registration
    self.creator.register("demo", "static.demo_nodes")
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 163, in register
    self._verify_identifier(package_identifier)
  File "/usr/share/miniconda3/envs/test/lib/python3.11/site-packages/pyiron_workflow/interfaces.py", line 212, in _verify_identifier
    raise ValueError(
ValueError: The package identifier is static.demo_nodes is not valid. Please ensure it is an importable module with a list of Node objects stored in the variable `nodes`.

```

So it looks like the problem is that `pympipool` requires to first CD and then discover tests. IMO that's a problem with `pympipool` and not with `pyiron_workflow`. I'll go over there and see if I can reproduce this behaviour more simply with some new workflows and get help there.

liamhuber:
Dammit, I [can't invoke the reusable workflow as a step](https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations). That means, and the docs explicitly point this case out, that I can't modify the `GITHUB_ENV` of the called workflow.

If this problem can't be fixed directly in `pympipool` that will mean adding a step like below to the reusable workflow itself.

```
    - name: "Add test dirs to pythonpath (pympipool compliance)"
      run: |
        PWD=$(pwd)
        echo "PYTHONPATH=$PWD/tests/benchmark:$PYTHONPATH" >> $GITHUB_ENV
        echo "PYTHONPATH=$PWD/tests/integration:$PYTHONPATH" >> $GITHUB_ENV
        echo "PYTHONPATH=$PWD/tests/unit:$PYTHONPATH" >> $GITHUB_ENV
```

liamhuber:
Per [pympipool #239](), I'm trying to modify the centralized CI to allow the python path to be expanded. Targeting the branch seems to have failed and push-pull didn't even run:

```
error parsing called workflow
".github/workflows/push-pull.yml"
-> "pyiron/actions/.github/workflows/push-pull-main.yml@tests_in_python_path"
: failed to fetch workflow: reference to workflow should be either a valid branch, tag, or commit
```

Ahaaaa, because I have a typo in "pythonpath"

liamhuber:
Perfect! Worked for all three OSs. I updated the upstream branch to also apply the same trick for the coverage job, which is still hanging.

liamhuber:
Note that once everything is settled, the workflow tags will need to be reset to main or a release.

Issue #72, opened 2023-11-07
:bulb: For-loop syntax?


liamhuber:
I want to migrate our discussion on for-loop syntax from the meeting discussion to its own issue, as it's getting a bit in-depth.

I had posted an example calculation that used the current (and not at all good) for-loop syntax:

```python
wf.ev = wf.create.meta.for_loop(
    loop_body_class=strained_energy,
    length=len(strains),  # I don't like that this is not dynamic
    iterate_on="strain"
)(
    name="Al", 
    STRAIN=strains.tolist()
)
```

@JNmpi replied:

> Please find below some syntax suggestions which I would like to have and which (at least to me) appear to be intuitive:
> 
> Replace 
> ```
> wf.ev = wf.create.meta.for_loop( loop_body_class=strained_energy, length=len(strains), # I don't like that this is not dynamic iterate_on="strain" )( name="Al", STRAIN=strains.tolist() )
> ```
> 
> by:
> 
> ```
>      wf.ev = strained_energy(name='Al', STRAIN=np.linspace(-0.05, 0.05, 5))
> ```
> 
> An alternative syntax would be:
> 
> ```
>      wf.ev = strained_energy(name='Al', strain__array=np.linspace(-0.05, 0.05, 5))
> ```
> 
> or
> 
> ```
>      wf.ev = strained_energy(name='Al', strain__list=np.linspace(-0.05, 0.05, 5))
> ```
> 
> Resolving the **kwargs argument the implementation should be rather straightforward. In one of my code examples (`iter_pool`) I implemented and used such a construction. If we later get type checking we could use the generic variable (strain) rather than one of the modified constructs (STRAIN, strain__array, etc.).

(Note that the registration aside is in its own issue, #71)

The @samwaseda replied:

> With @JNmpi 's suggestion, I fear that the subsequent node might not know whether to treat the array as one item or to loop over each element, especially in this example since we can set a float for a hydrostatic strain and an array for uniaxial strains in pyiron. I would probably suggest something like
> 
> ```python
> wf.ev = strained_energy(name='Al', STRAIN=wf.create.array(np.linspace(-0.05, 0.05, 5)))
> ```
> 
> I have to look into the actual code to see how its implementation could look...

liamhuber:
[...]

PR #78, opened 2023-11-15
Nodes for the `atomistics` package

jan-janssen:
Example:
```
from pyiron_workflow.atomistics_library.calculatornodes import get_emt, calc_with_calculator
from pyiron_workflow.atomistics_library.tasknodes import get_evcurve_task_generator, analyse_structures, generate_structures, get_bulk

wf = Workflow("evcurve")
wf.get_structure = get_bulk(element="Al")
wf.get_task_generator = get_evcurve_task_generator(structure=wf.get_structure)
wf.generate_structures = generate_structures(instance=wf.get_task_generator)
wf.get_calculator = get_emt()
wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures, calculator=wf.get_calculator)
wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)

wf.draw()

wf.run()
```
As the workflows in the `atomistics` package all follow the same pattern of (0) initializing the generator, (1) generating atomic structures, (2) evaluate the atomic structures with a simulation code and (3) analyse the results, the integration was rather straight forward.

I recently updated the `atomistics` package and its documentation https://atomistics.readthedocs.io/en/latest/ so this might be helpful to better understand the classes that I wrapped here.

### DIFF
--- a/pyiron_workflow/atomistics_library/macronodes.py
+++ b/pyiron_workflow/atomistics_library/macronodes.py
@@ -1,4 +1,5 @@
 from pyiron_workflow.macro import Macro, macro_node
+from pyiron_workflow.function import single_value_node
 from pyiron_workflow.atomistics_library.calculatornodes import calc_with_calculator
 from pyiron_workflow.atomistics_library.tasknodes import (
     get_elastic_matrix_task_generator,
@@ -9,12 +10,28 @@ from pyiron_workflow.atomistics_library.tasknodes import (
 )
 
 
+@single_value_node("instance")
+def get_instance(instance):
+    return instance
+
+
+@macro_node()
+def internal_macro(wf: Macro) -> None:
+    wf.get_instance = get_instance()
+    wf.generate_structures = generate_structures(instance=wf.get_instance)
+    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
+    wf.fit = analyse_structures(instance=wf.get_instance, output_dict=wf.calc_with_calculator)
+    wf.inputs_map = {
+        "get_instance__instance": "instance",
+        "calc_with_calculator__calculator": "calculator",
+    }
+    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+
+
 @macro_node()
 def get_energy_volume_curve(wf: Macro) -> None:
     wf.get_task_generator = get_evcurve_task_generator()
-    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
-    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
-    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
+    wf.internal = internal_macro(instance=wf.get_task_generator)
     wf.inputs_map = {
         "get_task_generator__structure": "structure",
         "get_task_generator__num_points": "num_points",
@@ -23,34 +40,30 @@ def get_energy_volume_curve(wf: Macro) -> None:
         "get_task_generator__vol_range": "vol_range",
         "get_task_generator__axes": "axes",
         "get_task_generator__strains": "strains",
-        "calc_with_calculator__calculator": "calculator",
+        "internal__calculator": "calculator",
     }
-    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+    wf.outputs_map = {"internal__fit_dict": "fit_dict"}
 
 
 @macro_node()
 def get_elastic_matrix(wf: Macro) -> None:
     wf.get_task_generator = get_elastic_matrix_task_generator()
-    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
-    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
-    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
+    wf.internal = internal_macro(instance=wf.get_task_generator)
     wf.inputs_map = {
         "get_task_generator__structure": "structure",
         "get_task_generator__num_of_point": "num_of_point",
         "get_task_generator__eps_range": "eps_range",
         "get_task_generator__sqrt_eta": "sqrt_eta",
         "get_task_generator__fit_order": "fit_order",
-        "calc_with_calculator__calculator": "calculator",
+        "internal__calculator": "calculator",
     }
-    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+    wf.outputs_map = {"internal__fit_dict": "fit_dict"}
 
 
 @macro_node()
 def get_phonons(wf: Macro) -> None:
     wf.get_task_generator = get_phonons_task_generator()
-    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
-    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
-    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
+    wf.internal = internal_macro(instance=wf.get_task_generator)
     wf.inputs_map = {
         "get_task_generator__structure": "structure",
         "get_task_generator__interaction_range": "interaction_range",
@@ -59,6 +72,6 @@ def get_phonons(wf: Macro) -> None:
         "get_task_generator__dos_mesh": "dos_mesh",
         "get_task_generator__primitive_matrix": "primitive_matrix",
         "get_task_generator__number_of_snapshots": "number_of_snapshots",
-        "calc_with_calculator__calculator": "calculator",
+        "internal__calculator": "calculator",
     }
-    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+    wf.outputs_map = {"internal__fit_dict": "fit_dict"}
--- a/pyiron_workflow/atomistics_library/macronodes.py
+++ b/pyiron_workflow/atomistics_library/macronodes.py
@@ -1,0 +1,64 @@
+from pyiron_workflow.macro import Macro, macro_node
+from pyiron_workflow.atomistics_library.calculatornodes import calc_with_calculator
+from pyiron_workflow.atomistics_library.tasknodes import (
+    get_elastic_matrix_task_generator,
+    get_evcurve_task_generator,
+    get_phonons_task_generator,
+    analyse_structures,
+    generate_structures
+)
+
+
+@macro_node()
+def get_energy_volume_curve(wf: Macro) -> None:
+    wf.get_task_generator = get_evcurve_task_generator()
+    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
+    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
+    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
+    wf.inputs_map = {
+        "get_task_generator__structure": "structure",
+        "get_task_generator__num_points": "num_points",
+        "get_task_generator__fit_type": "fit_type",
+        "get_task_generator__fit_order": "fit_order",
+        "get_task_generator__vol_range": "vol_range",
+        "get_task_generator__axes": "axes",
+        "get_task_generator__strains": "strains",
+        "calc_with_calculator__calculator": "calculator",
+    }
+    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+
+
+@macro_node()
+def get_elastic_matrix(wf: Macro) -> None:
+    wf.get_task_generator = get_elastic_matrix_task_generator()
+    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
+    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
+    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
+    wf.inputs_map = {
+        "get_task_generator__structure": "structure",
+        "get_task_generator__num_of_point": "num_of_point",
+        "get_task_generator__eps_range": "eps_range",
+        "get_task_generator__sqrt_eta": "sqrt_eta",
+        "get_task_generator__fit_order": "fit_order",
+        "calc_with_calculator__calculator": "calculator",
+    }
+    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
+
+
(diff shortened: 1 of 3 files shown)


liamhuber (approved):
Nice, LGTM. Good to see that it was so straightforward to wrap these, as that's absolutely the intent. 

Ultimately I want to spin _all_ the specific nodes out of this repo and into a different one, but this is a perfectly reasonable place for them to live right now.

Changes: 
- Please move `atomistics_library` to `node_library/atomistics`. 
- Then "nodes` in `calculatornodes` and `tasknodes` can all dropped, since these live in `node_library/atomistics/*` anyhow. (In the future you'll be able to register sub-domains with the creator class so that you can do something like `Workflow.register("atomistics", "pyiron_workflow.node_library.atomistics"); Workflow.create.atomistics.task.Foo`, but in the meantime at least it keeps all the specific nodes isolated from the raw infrastructure in a special folder)
- And `atomistics` (+version pin) can be added to the `extras_require["node_library" ]` in `setup.py`. (I'll add it to the notebooks env in a later PR as I work these into the educational material)

Nit: I would actually favour dropping `output_labels="..."` in the decorator and just pass the label without any keyword. IMO it's a little cleaner, and would make it more compatible with my intended change in the signature of the wrapper from `wrapper(output_labels: Optional[list[str] | str)` to `wrapper(*output_labels: str)`.

jan-janssen:
> * Please move `atomistics_library` to `node_library/atomistics`.

The issue with moving is that the module is named `atomistics` and the package I try to link is named `atomistics` as well, which leads to conflicts. In addition, I was wondering if there are any plans to store specific nodes separately, for example in this case the calculators which evaluate energy and forces for a given atomistic structure and the tasks which define what to do. 

> * Then "nodes`in`calculatornodes`and`tasknodes`can all dropped, since these live in`node_library/atomistics/*`anyhow. (In the future you'll be able to register sub-domains with the creator class so that you can do something like`Workflow.register("atomistics", "pyiron_workflow.node_library.atomistics"); Workflow.create.atomistics.task.Foo`, but in the meantime at least it keeps all the specific nodes isolated from the raw infrastructure in a special folder)

I guess this is where I am confused as there is already an `atomistics` module in the `node_library`. 

> * And `atomistics` (+version pin) can be added to the `extras_require["node_library" ]` in `setup.py`. (I'll add it to the notebooks env in a later PR as I work these into the educational material)

Does this mean we have one node library for everything? I thought it would be nice to have a node library for the `atomistics` package as in future this might become a standalone node library. 

> Nit: I would actually favour dropping `output_labels="..."` in the decorator and just pass the label without any keyword. IMO it's a little cleaner, and would make it more compatible with my intended change in the signature of the wrapper from `wrapper(output_labels: Optional[list[str] | str)` to `wrapper(*output_labels: str)`.

Sure, that should be quick. 

Some more questions from my side: 

* I would like to have the `calculator_nodes` be inputs for macros, but I was not exactly sure how I specify the node to be an input as it needs to connected to the evaluation node. 
* In addition, I would like to provide a set of `macros` one for each workflow like `calculated_energy_volume_curve()`, `calculate_elastic_matrix()` and so on. Would you store the macros in the same node library, or should the macros be moved to a different file, just to make it more clear for the users that by default I do not expect them to use individual nodes but rather recommend using the macros directly.

jan-janssen:
I added a series of macro nodes: 
```
wf = Workflow("evcurve")
wf.get_structure = get_bulk()
wf.get_calculator = get_emt()
wf.evcurve = evcurve(structure=wf.get_structure, calculator=wf.get_calculator)

wf.draw()
wf.run()
```

pyiron_workflow/atomistics_library/macronodes.py
```diff
@@ -0,0 +1,64 @@
...
+    generate_structures
+)
+
+
+@macro_node()
+def get_energy_volume_curve(wf: Macro) -> None:
+    wf.get_task_generator = get_evcurve_task_generator()
+    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
+    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
+    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
```
jan-janssen, line 17: These three nodes are required for every workflow, so I would like to create a macro like:
```
@macro_node()
def internal(wf: Macro) -> None:
    wf.generate_structures = generate_structures(instance=wf.get_task_generator)
    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
    wf.fit = analyse_structures(instance=wf.get_task_generator, output_dict=wf.calc_with_calculator)
    wf.inputs_map = {
        "generate_structures__instance": "instance",
        "analyse_structures__instance": "instance",
        "calc_with_calculator__calculator": "calculator",
    }
    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
```

The issue is that both `generate_structures()` and `analyse_structures()` should receive the same `instance` as input. Is this possible

liamhuber: It's possible to ensure that both `generate_structures()` and `analyse_structures()` get the same `instance` input, but not by leveraging the `inputs_map`. The `inputs/outputs_map` are (except for `None` which turns IO off) bijective maps. That means enforcing the same input needs to be done with graph connections rather than with the IO mapping. We could consider changing this expectation, but having the IO map be bijective has advantages that we'd be sacrificing for this convenience.

In the mean time, you can use a `standard.UserInput` node to take the `instance` input, and then connect its output (which is just plainly passed through the node)  to both targets:

```python
@macro_node()
def internal(wf: Macro) -> None:
    wf.instance = wf.create.standard.UserInput()
    wf.generate_structures = generate_structures(instance=wf.instance)
    wf.calc_with_calculator = calc_with_calculator(task_dict=wf.generate_structures)
    wf.fit = analyse_structures(output_dict=wf.calc_with_calculator, instance=wf.instance)
    wf.inputs_map = {
        "instance__user_input": "instance",
        "calc_with_calculator__calculator": "calculator",
    }
    wf.outputs_map = {"fit__fit_dict": "fit_dict"}
```

I imagine `UserInput` being useful once we have a GUI too, but this is actually an intended use of the design: mapping user input into the macro so that a single value can go multiple places. IMO the extra one line of inconvenience is worth it in order for the purity of the bijective mapping.

The only thing I _don't_ like is that I'm not sure how easy it is to type-hint this. It might be as simple as `wf.instance.inputs.user_input.type_hint = WhateverClass; wf.instance.outputs.user_input.type_hint = WhateverClass`, but I would need to dig in and see if the type checker is really storing raw types or if they get processed (e.g. `int | float` for sure gets converted to something like `typing.Union[int, float]`). The positive side is that such post-facto modifications of the type hinting are, in principle, possible and relatively straightforward, so it's just a matter of cooking up sexy syntax for it at a later date.

jan-janssen: I was able to solve this by defining a simple `get_instance()` node:
```
@single_value_node("instance")
def get_instance(instance):
    return instance
```

liamhuber:
> The issue with moving is that the module is named `atomistics` and the package I try to link is named `atomistics` as well, which leads to conflicts. 

Ahh, you are 110% right. Sorry. TBH, I care so little for the nodes currently in `atomistics` that I forgot what I called them...

Ok, so just forget moving them then. I still want to do it, and I think what is currently `atomistics` should be `pyiron_atomistics` so that this can be `atomistics`, but that will screw up the demo notebooks. I'll be responsible for making a PR later that moves this where I want it.

(Assuming, of course, that we deign to even keep the nodes currently in `atomistics` -- I rather suspect that between these node's and Joerg's nodes (#33), they are obsolete.)

> In addition, I was wondering if there are any plans to store specific nodes separately, for example in this case the calculators which evaluate energy and forces for a given atomistic structure and the tasks which define what to do.

I'm not sure I understand the contrast in your example, but there is definitely a plan to boot all the actual nodes out of this repo and into one or more node package repos -- except perhaps for the `standard` package that relies only on the standard library and provides basic things like an `If` node.

> I guess this is where I am confused as there is already an `atomistics` module in the `node_library`.

I guess this is resolved since the answer is just I was being dumb.

> 
> > * And `atomistics` (+version pin) can be added to the `extras_require["node_library" ]` in `setup.py`. (I'll add it to the notebooks env in a later PR as I work these into the educational material)
> 
> Does this mean we have one node library for everything? I thought it would be nice to have a node library for the `atomistics` package as in future this might become a standalone node library.

I 100% agree. In the medium-term (January? February?) I want to boot out (almost (see `standard` comment)) everything that is in `node_library` into one or more repos. IMO all the nodes you add here should be grouped together, but, as you say, it may be wise to have them separate from _other_ nodes.

One solution I can see is to have `atomistics/nodes/*.py` right in the `atomistics` package, and put `pyiron_workflow` as an optional dependency there. Another is to have `atomistics_nodes`, or `pyiron_nodes/atomistics` where `pyiron_nodes` has a variety of independent sub-modules. Honestly, I'm open to a number of things, but I want to keep the stuff here as bare-bones as possible in the long run, and that means kicking out domain-specific (or other-package-dependent) node definitions.

For the time being, I think it's ok to just keep dumping specific nodes into this repo until we figure out what we want.

> Sure, that should be quick.

👍 
 
> Some more questions from my side:
> 
> * I would like to have the `calculator_nodes` be inputs for macros, but I was not exactly sure how I specify the node to be an input as it needs to connected to the evaluation node.

I'm not sure I follow you here. In general, node instances are not valid IO to other instantiated nodes -- and I feel rather strongly about keeping it that way, as functional and idempotent behaviour goes straight to hell when you travel down that path. 

However, there are times when you do need variable node class as part of a macro definition. I've been calling these "meta"-nodes (cf. `pyiron_workflow.meta`), and the idea is that you dynamically define a new node _class_ that takes other node _classes_ as input, but when you actually instantiate it, it is building a particular macro with a particular class involved. #72 is good reading for current thoughts on this topic, particularly some syntactic sugar for skipping the middle class definition. 

I both hope and expect you have the tools already available to accomplish what you want, but I humbly recognize that the syntax is still in desparate need of some attention.

> * In addition, I would like to provide a set of `macros` one for each workflow like `calculated_energy_volume_curve()`, `calculate_elastic_matrix()` and so on. Would you store the macros in the same node library, or should the macros be moved to a different file, just to make it more clear for the users that by default I do not expect them to use individual nodes but rather recommend using the macros directly.

For now I would put them in a separate file, but as long as these macros only depend on the standard package and other nodes defined here, I would definitely expect these to all live in the same repo long-term. This is a good example of the benefit of sub-domains on the creator -- the first good one we've had so far I think. I.e. `wf.create.atomistics.calculator/task/macro` exactly as you're laying out the filenames now. (Or `science` or `material_properties` instead of `macro` or whatever, but you get the idea and I'm fine with `macro` for now if you are)

jan-janssen:
> Nice, LGTM. Good to see that it was so straightforward to wrap these, as that's absolutely the intent.

Wrapping functions as single value nodes was definitely easy, but I feel the creation of macros is still rather complex. As an example: 
```
@macro_node()
def get_elastic_matrix(wf: Macro) -> None:
    wf.get_task_generator = get_elastic_matrix_task_generator()
    wf.internal = internal_macro(instance=wf.get_task_generator)
    wf.inputs_map = {
        "get_task_generator__structure": "structure",
        "get_task_generator__num_of_point": "num_of_point",
        "get_task_generator__eps_range": "eps_range",
        "get_task_generator__sqrt_eta": "sqrt_eta",
        "get_task_generator__fit_order": "fit_order",
        "internal__calculator": "calculator",
    }
    wf.outputs_map = {"internal__fit_dict": "fit_dict"}
```
In practice I would rather do: 
```
@single_value_node("fit_dict")
def get_elastic_matrix(structure, calculator, num_of_point=5, eps_range=0.05, sqrt_eta=True, fit_order=2):
    return internal_macro(instance=get_elastic_matrix_task_generator(
        structure=structure, 
        calculator=calculator, 
        num_of_point=num_of_point, 
        eps_range=eps_range, 
        sqrt_eta=sqrt_eta, 
        fit_order=fit_order
    ))
```
The obvious limitation of the second variant is that I loose the ability to modify the internals of a function, but it is much easier to write. So as a consequence it would be interesting to modify the syntax of the macro definition in a way that it becomes more similar to the definition of a single_value_node.

liamhuber:
> Wrapping functions as single value nodes was definitely easy, but I feel the creation of macros is still rather complex. As an example:

I want to be careful to draw a distinction between "complex" and "verbose" here... do you really feel that this is actually complex? To me it's just a two-node workflow, which is pretty straightforward.

> In practice I would rather do:
> ...
> The obvious limitation of the second variant is that I loose the ability to modify the internals of a function, but it is much easier to write. So as a consequence it would be interesting to modify the syntax of the macro definition in a way that it becomes more similar to the definition of a single_value_node.

Yes, this is indeed the limitation. Not only do you lose the ability to modify them, you also lose the ability to query them post-facto, e.g. to see intermediate output.

`Function` nodes and `Composite` nodes are really, fundamentally different in my view. The advantages of a `Composite` structure are that they can be flexible (exposing/hiding new IO, replacing nodes to change behaviour) and transparent (intermediate values are available for investigation, flow can be easily drawn as a graph, etc). The disadvantage is that they are slower because they come with the overhead of everything we do in `pyiron_workflow` to get stuff running as a node.

There is no limit to the internal complexity of `Function` nodes and, in principle, any macro that consists only of `Function` nodes can be trivially re-written as a `Function` node -- in your case where you start with the imported functions this is obvious, but `Function.node_function` is actually a _class level attribute_, so these functions are available to glue together even for people who only have the nodes. And while it might be a bit of work for nested macros, because you'd need to go through and "functionalize" each layer of macros, this can thus be done for _any macro_. This is intentional design, so that if you have some workflow that you absolutely love and know is working perfectly, you're free to construct super-optimized and super-rigid versions of it.

But, ultimately, I don't _want_ to make macro definitions look like function node definitions -- I want them to look like workflow definitions, and indeed intend that most of them are just copy-pastes from people's workflows in a notebook. I do agree that the definitions of IO label maps is un-sexy, and am open to ideas for more sugary syntax there, but I still want all the `macro.foo = SomeNode(); ...` lines to look just like that.

I'll make a stacked PR demonstrating my currently preferred syntax for avoiding code duplication here.

liamhuber:
There's the macro PR.

The only other thing I really see missing is a node for converting the returned dictionary into independent channels. Honestly though, that's a shortcoming of `pyiron_workflow` and not the fault of `atomistics`. I won't get it done today, but I'll try to think this week of how to cleanly implement such a dict-to-many-channels node and add it to the standard package. No need for that to hold up merging here, it just will make it more useful in the future.

PR #79, opened 2023-11-19
Make py files node packages

liamhuber:
This is just so we can `Workflow.register` these files; another PR changing the macros will follow.


liamhuber:
@jan-janssen, since this is super minimal and strictly necessary for the registration, I'm going to merge it without review.

PR #80, opened 2023-11-19
Refactor macros

liamhuber:
Old: define a generic macro shared by all atomistics workflows leveraging task generators, then define a set of macros each of which instantiates the task generator node and the generic macro node, then connects them and manually re-maps all the task generator input to be available at the macro level.

New: Create a meta-workflow that takes the task generator node as input and returns a macro class in which all the task generator IO has been dynamically re-mapped to be available at the macro level; instantiate this meta-macro once per task generator to get a set of macros.

There is absolutely nothing wrong with the "old" version, it's a perfectly legitimate and legal way to construct a set of macros! Further, the biggest advantage of the "new" way -- that the task generator input is dynamically remapped -- could also be used right in the "old" way. Using a "meta-macro" is purely stylistic and my personal preference.

@jan-janssen, it's your node package so I think it's fair if you use whatever one you prefer. However, if you chose to stick with the implementation on the base branch, I would still suggest to leverage the same dynamic remapping just to shorten the node definitions.

```python

Workflow.register("macros", "pyiron_workflow.atomistics_library.macronodes")
Workflow.register("calculators", "pyiron_workflow.atomistics_library.calculatornodes")
Workflow.register("tasks", "pyiron_workflow.atomistics_library.tasknodes")

wf = Workflow("ev")
wf.bulk = wf.create.tasks.GetBulk("Al")
wf.calc = wf.create.calculators.GetEmt()
# wf.old_ev_curve = wf.create.macros.GetEnergyVolumeCurve(
#     structure=wf.bulk,
#     calculator=wf.calc,
# )
wf.new_ev_curve = wf.create.macros.EnergyVolumeCurve(
    structure=wf.bulk,
    calculator=wf.calc,
)
out = wf()

# out.old_ev_curve__internal__fit__result_dict["poly_fit"] - out.new_ev_curve__result_dict["poly_fit"]
```

The commented out stuff works on the first commit of this PR, where both old and new exist at once. The output is just an array of zeros, as expected and desired.

I didn't do the same sort of thorough test for the other two, but I did make sure that they instantiated OK and had nice clean IO:

```python
p = Workflow.create.macros.Phonons()
p.inputs.labels, p.outputs.labels
```

```
(['structure',
  'interaction_range',
  'factor',
  'displacement',
  'dos_mesh',
  'primitive_matrix',
  'number_of_snapshots',
  'calculator'],
 ['result_dict'])
```

and 

```python
l = Workflow.create.macros.ElasticMatrix()
l.inputs.labels, l.outputs.labels
```

```
(['structure',
  'num_points',
  'fit_type',
  'fit_order',
  'vol_range',
  'axes',
  'strains',
  'calculator'],
 ['internal__fit__result_dict'])
```

### DIFF
--- a/pyiron_workflow/atomistics_library/macronodes.py
+++ b/pyiron_workflow/atomistics_library/macronodes.py
@@ -10,7 +10,6 @@ from pyiron_workflow.atomistics_library.tasknodes import (
 
 
 def atomistics_meta_macro(task_generator_node_class, macro_name) -> type[Macro]:
-
     def generic_macro(wf: Macro) -> None:
         wf.tasks = task_generator_node_class()
         wf.structures = generate_structures(instance=wf.tasks)
@@ -18,11 +17,13 @@ def atomistics_meta_macro(task_generator_node_class, macro_name) -> type[Macro]:
         wf.fit = analyse_structures(instance=wf.tasks, output_dict=wf.calc)
         inputs_map = {
             # Dynamically expose _all_ task generator input directly on the macro
-            "tasks__" + s: s for s in wf.tasks.inputs.labels
+            "tasks__" + s: s
+            for s in wf.tasks.inputs.labels
         }
         inputs_map["calc__calculator"] = "calculator"
         wf.inputs_map = inputs_map
         wf.outputs_map = {"fit__result_dict": "result_dict"}
+
     generic_macro.__name__ = macro_name
 
     return macro_node()(generic_macro)
@@ -34,7 +35,8 @@ elastic_matrix = atomistics_meta_macro(
 
 
 energy_volume_curve = atomistics_meta_macro(
-    get_evcurve_task_generator, "energy_volume_curve",
+    get_evcurve_task_generator,
+    "energy_volume_curve",
 )
 
 phonons = atomistics_meta_macro(get_phonons_task_generator, "phonons")
--- a/pyiron_workflow/atomistics_library/macronodes.py
+++ b/pyiron_workflow/atomistics_library/macronodes.py
@@ -1,5 +1,4 @@
 from pyiron_workflow.macro import Macro, macro_node
-from pyiron_workflow.function import single_value_node
 from pyiron_workflow.atomistics_library.calculatornodes import calc_with_calculator
 from pyiron_workflow.atomistics_library.tasknodes import (
     get_elastic_matrix_task_generator,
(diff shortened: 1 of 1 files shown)


pyiron_workflow/atomistics_library/macronodes.py
```diff
@@ -10,77 +9,41 @@
...
+        wf.calc = calc_with_calculator(task_dict=wf.structures)
+        wf.fit = analyse_structures(instance=wf.tasks, output_dict=wf.calc)
+        inputs_map = {
+            # Dynamically expose _all_ task generator input directly on the macro
+            "tasks__" + s: s
+            for s in wf.tasks.inputs.labels
+        }
+        inputs_map["calc__calculator"] = "calculator"
+        wf.inputs_map = inputs_map
+        wf.outputs_map = {"fit__result_dict": "result_dict"}
```
jan-janssen, line 25: Can this part be done more automatically? I mean can the workflow / macro object `wf` return a list of all input nodes which are not connected and all output nodes which are not connected and then just connect these to the nodes of the macro? While I understand this is not the behaviour everybody wants, I could see myself using this kind of functionality for a wide range of macros.

liamhuber: This is _exactly_ what's done, but by default the macro-level channel labels are "{node_name}__{channel_name}" to make sure nothing conflicts and to reduce ambiguity.

liamhuber: (well, not "exactly" as I-I and O-O are not "connections", connections are only for I-O/O-I, but anyhow the links are created to every child IO that is otherwise unconnected).

liamhuber: I've been thinking about this in the back of my head during the meeting, and what I could see being a good compromise is offering some sort of keyword argument like `flatten_io_labels` or similar that tries to do this flattening automatically. I would still be more comfortable if it only happens when the user asks for it, but I agree it should be more convenient to do so than manually adding the snippet I use here.

jan-janssen: I really like the idea of having a special keyword like `flatten_io_labels` that would simplify it a lot.

liamhuber: Yeah, I'm really sold on the idea now too. Are you comfortable merging this with the bespoke implementation here? I'm not sure how fast I'll get to implementing it a a generic feature, so I would be happy to have this one-off implementation as an example of how to do it right away, then come back and replace this with the keyword once the generic version exists.

jan-janssen (approved):
Looks good to me

PR #91, opened 2023-11-27
Jnmpi lammps nodes patch 0.1.0

liamhuber:
Almost no changes were needed to get the lammps nodes branch compliant with 0.1.0, just a couple of parameter changes from int to float to get values compliant with a type hint.


liamhuber:
Change for this were super minor, just making ints into floats to comply with type hinting, so I'm going to merge without review

Issue #38, opened 2023-10-19
Example for experimental workflows


SteffenBrinckmann:
Joerg and Sarath introduced me to the workflow engine. Very nice work and useful. Below are some feedback that you might consider along with my example

- I find the requirements a little much: if I only install workflows, I don't understand why I need e.g. SQL stuff
- can the workflow.run parameters become easier
``` python
wf.step.inputs.name  = 'sem'              #step 3
wf.step.inputs.param = {'voltage':'30'}
out = wf.run()
```
could become one line: 
``` python
out = wf.run( {'step':{name:'sem', 'param':{'voltage':30}}}) 
```
- the out is also not beautiful (see below) because I need to return 3 values, ... have to pack them into a list, ...
``` python
print('Output filename',out['step__[sample, fileName, metadata]'][1])
```
- a draw function that is build on matplotlib as I don't use jupyter but a conventional python script.

Full example (I can supply the Storage and Sample classes on request)
``` python
from pathlib import Path
from pyiron_workflow.workflow import Workflow
from experimental_workflows.storage import Storage, step
from experimental_workflows.sample import Sample

wf = Workflow('example_workflow')         #name
folder = Path(__file__).parent.parent/'procedures'
wf.step = step(storage=Storage(folder))   #define step and link to storage for procedures

sample = Sample('FeAl')
wf.step.inputs.sample = sample            #since only use one sample, define only once

wf.step.inputs.name  = 'polish'           #step 1
out = wf.run()

wf.step.inputs.name  = 'light microscopy' #step 2
out = wf.run()
print('Output filename',out['step__[sample, fileName, metadata]'][1])

wf.step.inputs.name  = 'sem'              #step 3
wf.step.inputs.param = {'voltage':'30'}
out = wf.run()
print('Output filename',out['step__[sample, fileName, metadata]'][1])
```

liamhuber:
Hey @SteffenBrinckmann, 

Glad to hear you're working with it and thanks for the feedback!

> I find the requirements a little much: if I only install workflows, I don't understand why I need e.g. SQL stuff

💯

We currently depend directly on `pyiron_base`, which has a bunch of these dependencies, but we're really only using a couple of simple utility snippets that can be pulled out.
We also depend on `pyiron_atomistics` (and other modules) and that entire dependency stack, but this is only for the specific "node packages". The idea is to (a) move imports inside the node definitions themselves, making the dependency optional, and (b) eventually move the node packages off to their own repository(ies).

[...]

PR #93, opened 2023-11-27
Expose dict values as list

### LINKED ISSUE #40: Results to list easier
From #38 

Add `.to_list()` methods to `DataIO` and `DotDict` for getting results as lists instead of dicts easier. (`DotDict` does already have `.values()` but this is not a true list.)

liamhuber:
Just adds `to_list()` methods to `DotDict` and `DataIO` so the dictionaries can be case to value lists. The order is explicitly not guaranteed, but it winds up being the order items are added to the underlying dict.

As requested in #38.

Closes #40.

### DIFF
--- a/pyiron_workflow/io.py
+++ b/pyiron_workflow/io.py
@@ -171,6 +171,10 @@ class DataIO(IO, ABC):
     def to_value_dict(self):
         return {label: channel.value for label, channel in self.channel_dict.items()}
 
+    def to_list(self):
+        """A list of channel values (order not guaranteed)"""
+        return list(channel.value for channel in self.channel_dict.values())
+
     @property
     def ready(self):
         return all([c.ready for c in self])
--- a/pyiron_workflow/util.py
+++ b/pyiron_workflow/util.py
@@ -20,6 +20,10 @@ class DotDict(dict):
         for k, v in state.items():
             self.__dict__[k] = v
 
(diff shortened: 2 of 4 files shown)


liamhuber:
```
ERROR: test_connection (unit.test_io.TestDataIO.test_connection)
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/test_io.py", line 68, in test_connection
    self.input.x = self.input.y
    ^^^^^^^^^^^^
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/pyiron_workflow/io.py", line 76, in __setattr__
    self._assign_value_to_existing_channel(self.channel_dict[key], value)
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/pyiron_workflow/io.py", line 92, in _assign_value_to_existing_channel
    channel.connect(value.channel)
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/pyiron_workflow/channels.py", line 126, in connect
    raise ChannelConnectionError(
pyiron_workflow.channels.ChannelConnectionError: x (InputData) and y (InputData) share a generic type but were not a valid connection. Check channel classes, type hints, etc.
```

Riiight. Because the two tests each had the same name, only one of them was getting run and it wasn't this one. This actually does break expectations -- trying to assign input to input should raise an error.

PR #98, opened 2023-12-01
Full parallelism

### LINKED ISSUE #74: Let parallel nodes automatically execute in parallel
By breaking the `run` signal to `or_run` and `and_run`, where the former triggers a run if it receives a signal from _any_ of its connections, while the latter triggers after receiving at least one signal from _all_ its connections.

This can be used to resolve the inefficiency with our current DAG automation when combined with executors. The current DAG automation just builds a single linear chain of execution, so if the nodes could execute in parallel we totally lose any advantages there. The topological analysis already gives us "layers" of independent nodes, we're just missing the `Signals` infrastructure to take advantage of that! It hasn't mattered since executor support was so trivial, but now that `pympipool` seems to be working, it's time to fix this.

liamhuber:
Add a new signal `accumulate_and_run` to each node's signals, which leverages a new `AccumulatingInputSignal` that only fires after receiving at least one ping from _each_ of its upstream connections.

This gets new sexy syntax to complement the existing `run_first > run_second > run_third` syntax for regular `run` signal: `my_node << upstream_node1, upstream_node2, also_by_signal.signals.output.ran`.

`Composite` now leverages this new connection style in the automated execution flows for DAGs, so sending stuff to executors will actually give you speedup now, as long as the serialization time is short compared to the execution time.

Here's a little example of the speedup:

```python
from time import perf_counter, sleep

from pyiron_workflow.channels import NotData

@Workflow.wrap_as.single_value_node()
def Wait(t):
    sleep(t)
    return True
    
t = 2

wf = Workflow("serial")
wf.a = Wait(t)
wf.b = Wait(t)
wf.c = Wait(t)
wf.d = wf.create.standard.UserInput(t)
wf.automate_execution = False
wf.a > wf.b > wf.c > wf.d
wf.starting_nodes = [wf.a]
t0 = perf_counter()
out = wf()
while wf.outputs.d__user_input.value is NotData:
    sleep(0.001)
dt_serial = perf_counter() - t0

print(dt_serial)
>>>  6.008391406998271

wf = Workflow("parallel")
wf.a = Wait(t)
wf.b = Wait(t)
wf.c = Wait(t)
wf.d = wf.create.standard.UserInput(t)
wf.automate_execution = False
wf.d << wf.a, wf.b, wf.c
wf.starting_nodes = [wf.a, wf.b, wf.c]


with wf.create.Executor(max_workers=3, cores_per_worker=1) as executor:
    wf.a.executor = executor
    wf.b.executor = executor
    wf.c.executor = executor

    t1 = perf_counter()
    out = wf()
    while wf.outputs.d__user_input.value is NotData:
        sleep(0.001)
    dt_parallel = perf_counter() - t1

print(dt_parallel)
>>> 2.4772168039926328
```

So you can see that at least for nodes that are not pricey to serialize, O(1s) is already plenty slow enough to benefit from the parallelization.

Closes #74


liamhuber:
The pip check failure is some nonsense upstream and nothing to do with this PR:

```
tables 3.9.2 requires blosc2, which is not installed.
```

Issue #46, opened 2023-10-23
`pyiron_workflow` depends on `pyiron_atomistics` and `pyiron_base` leading to a large number of dependencies


srmnitc:
**Summary**

`pyiron_workflow` depends on `pyiron_atomistics` and `pyiron_base`, which in turn depends on a large number of packages. This makes the workflow package rather heavy for a user who might want to use it solely to leverage the power of workflows. Would it make sense to release the dependencies and make this package rather light weight?

liamhuber:
Absolutely. There's a clear path for this too, it just hasn't been high priority.

Our dependency on `pyiron_base` is currently just `logger` and `Singleton`. These could  be git-copied over here, or (better IMO) we can start a new repo of standard-libarary-only-depenency pyiron tools as discussed in [this issue over on base](https://github.com/pyiron/pyiron_base/issues/1208).

Our dependency on `pyiron_atomistics` is only for actual individual nodes, and not in the infrastructure. At the latest when the `node_library` submodule gets pulled out/broken apart into node library module(s), we'll lose this dependency. We can probably make it optional even before then by moving the import _inside_ the individual node functionalities.

liamhuber:
This was resolved and both those packages are out of the dependency stack, buuuut, I'm probably about to add base back in in order to get storage rolling, so I'll leave this open.

liamhuber:
> This was resolved and both those packages are out of the dependency stack, buuuut, I'm probably about to add base back in in order to get storage rolling, so I'll leave this open.

Nope, we'll just move `tinybase` here, per #161. So this is out, the dependency is strictly for the node libraries that depend on those modules.

liamhuber:
#160 introduces `pyiron_contrib` dependency, although it's not planned to stay this way. I'm re-opening to keep track of the shortcoming.

liamhuber:
I'm also slapping on `boto3` as long as [contrib is not playing nice](https://github.com/pyiron/pyiron_contrib/issues/551)

PR #109, opened 2023-12-05
Move generic stuff into a new submodule

liamhuber:
This doesn't get all the way to releasing a new `pyiron_snippets` package, but it's a first step. 

I moved my `pyiron_base` imports (including their git history) into a local submodule, along with the base test case class that helps with docs testing. Then I moved everything in `pyiron_workflow.utils` in there too. The code and tests for these generic things (which depend only on the standard library) are thus encapsulated in two directories and ready to be shipped off to their own repo. In the meantime, this also lets me remove my dependency on `pyiron_base`.


liamhuber:
This is also most of the weight for #46, since we already booted `pyiron_atomistics` into just the node library. Matplotlib is still left, but that's just a couple colour converter functions and should be easy to deal with in the new `snippets.colors` submodule.

Issue #107, opened 2023-12-04
Test docstrings in CI


liamhuber:
Using [`doctest`](https://docs.python.org/3/library/doctest.html)

PR #112, opened 2023-12-06
Test docs

liamhuber:
Using doctest. Docstrings in the unit tests, and the README in the integration tests.

I don't actually wind up using the `PyironTestCase` snippet, but I'm content to leave it around for now.


liamhuber:
Closes #107

PR #147, opened 2024-01-04
Registration

liamhuber:
Ahead of saving and loading workflows, we need the saved nodes to be able to know what node package they came from, so that workflows can register these on load, re-create the nodes, then the nodes can load their state from the data. (Note that Macros don't have the same problem, because they know how to create all their children and do so at initialization, so they can simply read off their saved states -- there is an edge case where a macro has `replace`d one or more of it's children, but we can deal with that later).

In principle each node package should have a unique identifier, but in practice I'm still just using the import module, e.g. `"pyiron_workflow.node_library.standard"`. This "identifier" now gets stored on the node classes when they are registered as a package, so they'll be available to save later.

Otherwise there's a bunch of refactoring, including moving more responsibility onto `NodePackage` and making the (for now) equivalence of the package identifier and module name more clear.

There's also a bit of extra syntactic sugar so that you can register stuff to a deeply nested domain from the get-go, e.g. `Workflow.create.register("my.deep.domain", "pyiron_workflow.node_library.standard"); Workflow.create.my.deep.domain.Add()`.


liamhuber:
Notebook failure appears to just be a versioning inconsistency with pyiron/atomistics; I'll fix it before merging but it doesn't impact the meat here

liamhuber:
Original notebook error seems to have been a caching error and outdated version of pyiron/atomistics. Now pyiron_atomistics is breaking in a strange (and still seemingly unrelated) way:
```
File /usr/share/miniconda3/envs/my-env/lib/python3.11/site-packages/pyiron_base/jobs/job/runfunction.py:700, in handle_failed_job(job, error)
    698     if job.server.run_mode.non_modal:
    699         state.database.close_connection()
--> 700     raise RuntimeError("Job aborted")
    701 else:
    702     return True, out

RuntimeError: Job aborted
```

liamhuber:
One problem was that updates to the `extras_require` in `setup.py` are not being automatically propagated to `.ci_support/notebooks-env.yml`. This is a CI bug, and I opened an issue for it: #148

liamhuber:
Looking for: ['jupyter', 'papermill', 'coveralls', 'coverage', 'bidict=0.22.1', 'cloudpickle=3.0.0', 'graphviz=8.1.0', 'matplotlib=3.8.2', 'pympipool=0.7.9', 'python-graphviz=0.20.1', 'toposort=1.10', 'typeguard=4.1.5', 'ase=3.22.1', 'atomistics=0.1.20', 'lammps', 'phonopy=2.21.0', 'pyiron_atomistics=0.4.1', 'pyiron-data=0.0.24', 'numpy=1.26.2']


├─ atomistics 0.1.20**  is requested and can be installed;
└─ pyiron_atomistics 0.4.1**  is uninstallable because it requires
   └─ atomistics >=0.1.12,<=0.1.18 , which conflicts with any installable versions previously reported.
```

Right, the dependabot upgrade of `atomistics` to 0.1.19 got through the tests before because it only impacts the notebooks test, but changes to setup.py by dependabot were not getting propagated to the actual notebooks env so the conflict never arose. I'm going to go patch this in a separate PR.

liamhuber:
Closes #152

PR #149, opened 2024-01-05
Notebooks env patch

liamhuber:
And updating the atomistics notebook to use the new API.

Basically a manual version of what should be automated, as discussed in #148

PR #150, opened 2024-01-05
Don't remove a readme that isn't there

liamhuber:


liamhuber:
I'm still getting the same problem as in #147 and #149, where the first time I actually try to run a lammps job it crashes. Since `pyiron_atomistics` imposes no upper bound on `pyiron_base`, I wind up getting `pyiron_base=0.6.20` together with the very old `pyiron_atomistics=0.3.5`. Let's try pinning base to the expected value [from pyiron_atomistics](https://github.com/pyiron/pyiron_atomistics/blob/de5068dd8bbcdc98bfd9a36f474ebca701342b8b/.ci_support/environment.yml#L19C3-L19C21)

liamhuber:
Nope. I honestly have no idea.

PR #157, opened 2024-01-10
Working dir emptiness check

liamhuber:
Just very minor convenience changes to the filesystem stuff: adds a property `DirectoryObject.is_empty` that checks agains length, and  modifies deletion so you can choose to only delete if it's empty: `DirectoryObject.(self, only_if_empty: bool = False)`

### DIFF
--- a/pyiron_workflow/files.py
+++ b/pyiron_workflow/files.py
@@ -43,8 +43,9 @@ class DirectoryObject:
     def create(self):
         self.path.mkdir(parents=True, exist_ok=True)
 
-    def delete(self):
-        delete_files_and_directories_recursively(self.path)
+    def delete(self, only_if_empty: bool = False):
+        if self.is_empty or not only_if_empty:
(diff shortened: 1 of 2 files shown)


pyiron_workflow/files.py
```diff
@@ -71,6 +72,10 @@ def create_subdirectory(self, path):
     def create_file(self, file_name):
         return FileObject(file_name, self)
 
+    @property
+    def is_empty(self) -> bool:
```
samwaseda, line 76: I'm not really sure, but isn't usually `is_xyz` a function?

liamhuber: Hmm, to my mind it was a property, while `empty` is a method since it's ambiguous since it might be a verb.

Ah, you're right though, [pathlib.Path](https://docs.python.org/3/library/pathlib.html) makes all it's `is_*` things methods. I'll change it.

samwaseda (approved):

PR #159, opened 2024-01-10
Add a shortcut for removing file(s) by name

liamhuber:
Introduces the following convenience method to `DirectoryObject`:

```python
    def remove(self, *files: str):
        for file in files:
            path = self.get_path(file)
            if path.is_file():
                path.unlink()
```

### DIFF
--- a/pyiron_workflow/snippets/files.py
+++ b/pyiron_workflow/snippets/files.py
@@ -75,6 +75,12 @@ class DirectoryObject:
     def is_empty(self) -> bool:
         return len(self) == 0
 
+    def remove(self, *files: str):
+        for file in files:
+            path = self.get_path(file)
+            if path.is_file():
+                path.unlink()
+
 
 class FileObject:
     def __init__(self, file_name: str, directory: DirectoryObject):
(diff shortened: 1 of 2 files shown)


pyiron_workflow/snippets/files.py
```diff
@@ -75,6 +75,12 @@ def create_file(self, file_name):
     def is_empty(self) -> bool:
         return len(self) == 0
 
+    def remove(self, *files: str):
```
samwaseda, line 78: If this is a function that deletes only files, then I would also call the function `remove_files`, because otherwise the distinction between `remove` and `delete` isn't so clear. If not, it at least merits a doc string that makes clear what's the difference.

samwaseda:
Actually I don't really get the intention of this PR. I think by default people should delete the file from `FileObject` and not from the directory (except when all files should be erased)

liamhuber:
Wouldn't that involve importing `FoleObject` and creating N of these based on the strings used in this shortcut and the `DirectoryObject`? This isn't horrid, I just wanted something even shorter. You can see the use in `Node.__post__` over in the storage PR

liamhuber:
@samwaseda if, on reflection, you think this is useful, feel free to merge it. However, I came around to your perspective that importing and instantiating a `FileObject` is also perfectly acceptable, and in my case (since I have only the one file I care about and already import `DirectoryObject`) not particularly more work. I've dropped it from the history of #160 and don't depend on it.

So if you still dislike it, just close the PR; I still don't hate it, but am quite indifferent now.

samwaseda:
Still not really sure about this one, but I also don't feel strongly against it, so I merged it :D

PR #174, opened 2024-01-24
Callbacks as strings

liamhuber:
This is stacked on top of #160, but looks more complicated than it is because it also merged in #173. Once #173 is merged, just the 4 commits from 23 Jan will show here.

Basically we add the requirement that callback functions stored on `InputSignal` need to be methods of the node that owns them. In practice this has always been the case, so there is no real change, and since you can put whatever method you want on 

Further, the requirement that callbacks be `arg`-free was already stated, but now we explicitly test for it and fail hard if it's violated.

Since the callback is now guaranteed to be a method of the owning `InputSignal.node` attribute, we can safely store its name in our state (a mere string) instead of storing the actual function object. This will make storage easier later.

### DIFF
--- a/pyiron_workflow/channels.py
+++ b/pyiron_workflow/channels.py
@@ -804,13 +804,17 @@ class InputSignal(SignalChannel):
 
     @staticmethod
     def _no_positional_args(func):
-        return sum(
-            1 for parameter in inspect.signature(func).parameters.values()
-            if (
-                parameter.default == inspect.Parameter.empty
-                and parameter.kind != inspect._ParameterKind.VAR_KEYWORD
+        return (
+            sum(
+                1
+                for parameter in inspect.signature(func).parameters.values()
+                if (
+                    parameter.default == inspect.Parameter.empty
+                    and parameter.kind != inspect._ParameterKind.VAR_KEYWORD
+                )
             )
-        ) == 0
+            == 0
+        )
 
     @property
     def callback(self) -> callable:
--- a/pyiron_workflow/channels.py
+++ b/pyiron_workflow/channels.py
@@ -464,7 +464,7 @@ class DataChannel(Channel, ABC):
         storage["value"] = self.value
 
     def from_storage(self, storage):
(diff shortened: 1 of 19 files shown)


pyiron_workflow/channels.py
```diff
@@ -777,7 +783,42 @@ def __init__(
...
+            sum(
+                1
+                for parameter in inspect.signature(func).parameters.values()
+                if (
+                    parameter.default == inspect.Parameter.empty
+                    and parameter.kind != inspect._ParameterKind.VAR_KEYWORD
+                )
+            )
+            == 0
+        )
```
samwaseda, line 817:
```suggestion
        for parameter in inspect.signature(func).parameters.values():
            if parameter.default == inspect.Parameter.empty and parameter.kind != inspect._ParameterKind.VAR_KEYWORD:
                return False
        return True
```

samwaseda (approved):
Looks good to me (except for just one minor comment)

pyiron_workflow/channels.py
liamhuber: Nice improvement-- Early stopping and more readable. Looking at this, I think we can do it with an `all()` comprehension and further get rid of the branching return!

liamhuber: I can commit this (so you're in the history here) and do that tomorrow, but if you see this in time feel free to modify the suggestion and directly commit it. (Assuming you like the idea)

samwaseda: Ah I see. I like the idea

samwaseda, line 817:
```suggestion
        return all([
            parameter.default != inspect.Parameter.empty
            or parameter.kind == inspect.Parameter.VAR_KEYWORD
            for parameter in inspect.signature(func).parameters.values()
        ])
```

samwaseda: I changed it silently, but apparently `inspect._ParameterKind.VAR_KEYWORD` is the same as `inspect.Parameter.VAR_KEYWORD`

PR #176, opened 2024-01-24
Copy and update __dict__

liamhuber:
Sometimes we mess with stuff in __getstate__, e.g. to avoid reflexive relationships between nodes and channels, but getting the state should never modify the current state, so always make a fresh dictionary. Similarly, the returned state might possibly not have all the content our current state does, so use update instead of overwriting.


liamhuber:
Oops, forgot to branch first. Will force-push the original branch

PR #179, opened 2024-01-25
`NotData` as singleton instead

liamhuber:
The original motivation was to get the non-data data working better with `h5io`, but I actually like the singleton better than using a raw class. Also the `is` comparisons all keep working, since it's literally the same instance.

Defining `__reduce__` allows pickle to get back the exact same global instance defined alongside the singleton class.

After @samwaseda's addition (#185), I start to wonder if this is actually something that would be nice to have generically available in `snippets`?

### DIFF
--- a/notebooks/deepdive.ipynb
+++ b/notebooks/deepdive.ipynb
@@ -90,7 +90,7 @@
    "id": "22ee2a49-47d1-4cec-bb25-8441ea01faf7",
    "metadata": {},
    "source": [
-    "The output is still empty (`NotData`) because we haven't `run()` the node:"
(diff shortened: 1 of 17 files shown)


samwaseda:
> After @samwaseda's addition (#185), I start to wonder if this is actually something that would be nice to have generically available in `snippets`?

Yes, I was about to post this comment, but then also thought since it's only used in the channel and it's anyway not visible to the user, I think we can postpone the discussion for now. As soon as we see a potential merit to have it in snippets, we can easily move it.

samwaseda (approved):
Looks good to me.

PR #180, opened 2024-01-25
Parent decorated nodes to the module their function came from

liamhuber:
Instead of `abc`.

This isn't particularly important by itself, just aesthetically nice,  but together with the two PRs I opened on `h5io` it lets us save and load with that package since the nodes become pickle-findable.

E.g. we can do this and it works*:

```python

wf = Workflow("some_graph")

wf.inp = wf.create.standard.UserInput(42)
wf.double = 2 * wf.inp
wf.out = wf.create.standard.UserInput(wf.double)

out = wf()

import h5io

h5io.write_hdf5(
    fname="stuff.h5",
    data=wf,
    use_state=True,
    overwrite=True,
)

wf_reloaded = h5io.read_hdf5(fname="stuff.h5")
```

*Kind of. `h5io` still borks for some node content, e.g. for the `atomistics` library it complains about some random method on the `Calculator` object that `EMT` inherits from. In contrast, `tinybase` can serialize atomistics workflows without any trouble -- I think because it ultimately falls back on `cloudpickle` if it gets really stuck?

### DIFF
--- a/pyiron_workflow/function.py
+++ b/pyiron_workflow/function.py
@@ -764,6 +764,7 @@ def _wrapper_factory(
                 ),
                 "node_function": staticmethod(node_function),
                 "_type_hints": get_type_hints(node_function),
+                "__module__": node_function.__module__,
             },
         )
 
--- a/pyiron_workflow/macro.py
+++ b/pyiron_workflow/macro.py
@@ -625,6 +625,7 @@ def macro_node(*output_labels, **node_class_kwargs):
                     **node_class_kwargs,
                 ),
                 "graph_creator": staticmethod(graph_creator),
+                "__module__": graph_creator.__module__,
             },
         )
 


liamhuber:
There's no tests yet because this is worth doing on purely aesthetic grounds for people who look at the `type()` of their nodes, but so far we don't use this for anything mission critical or promise any behaviour around it.

liamhuber:
> I tried a few things but the example code gives this error message:
> 
> ```python
> File ~/conda/lib/python3.11/site-packages/h5io/_h5io.py:434, in _triage_write(key, value, root, comp_kw, where, cleanup_data, slash, title, use_json, use_state)
>     431     return
>     433 err_str = "unsupported type %s (in %s)" % (type(value), where)
> --> 434 raise TypeError(err_str)
> 
> TypeError: unsupported type <class 'pyiron_workflow.channels.InputSignal'> (in <class 'pyiron_workflow.workflow.Workflow'>["run"])
> ```

Ah yeah, from here on out things will stop working on your machine unless you clone `h5io`, make a new branch, and merge all my PRs over there into it.

liamhuber:
@samwaseda, my original comment was worded so poorly as to be misleading, sorry. To actually see the effect of the PR you can just use 

```python
>>> from pyiron_workflow import Workflow
>>> 
>>> wf = Workflow("some_graph")
>>> 
>>> wf.inp = wf.create.standard.UserInput(42)
>>> type(wf.inp)
pyiron_workflow.node_library.standard.UserInput
```

Where before it was `abc.UserInput`

liamhuber:
> [liamhuber](https://github.com/liamhuber) requested a review from [samwaseda](https://github.com/samwaseda)

Actually, I'm going to compress the whole stack down into the main PR today, so that request was ill thought out

PR #185, opened 2024-01-29
`NOT_DATA` should work like None in the if-statement

samwaseda:
`NOT_DATA` should work like `None`, and therefore should be `else` in the `if`-statement.


liamhuber (approved):
Good thinking!

liamhuber:
The whole stack needs to pass the CI before merging into main, but this passes tests on my local machine so I'm going to merge it into the parent branch then propagate the changes up the remaining stack 🚀

Issue #182, opened 2024-01-26
Thoughts on storage


liamhuber:
This is just my stream-of-consciousness notes immediately post-meeting with @pmrv.
I'll come back and update it, and split it into two or more sensible actionable issues next week when I have more time. 
Meanwhile, I wanted it available in light of my request (#181) to meet with both @pmrv and @jan-janssen so we can move towards a storage solution that satisfies everyone.

## (De)serialization for remote processes

- Change the bit of `run` that gets shipped off to other processes to _include_ info about serializing the output
  - We'll probably also need to send off the fully scoped working directory
  - For performance, we'd ship out the run function and inputs, and ship back the outputs (not all of `self`)
- Serialize to some temp file, then at the last minute move the file to the "expected" location
- One the main process side, update your output, and either delete the temp
- When loading a graph, if you find a node who deserializes with a `running` state, go look for the "expected" save file; if you find it, deserialize it and use this output data in your regular callback (including a full (not just output) save (if requested), a checkpointing save (if requested), and firing output signals)
- In this way submitting to an `Executor` process or a queue process should look very similar
  - With the exception that what gets sent to the queue process may also be instructions for using an `Executor` nested inside that process

This should provide compatibility with both queues and restarting partially executed runs!

## Storage implementation

The second topic was storage details.
One thing I forgot to mention that I liked about `tinybase`'s storage solution was that it has a generic abstraction that would allow us to plug in non-hdf5 back ends.
Marvin was very amenable to the idea of using `__get/setstate__` natively (where it exists) to do the `re/store` process, so that we don't necessarily need to define those custom functions.
The rough idea now is to use `__getstate__` recursively to get key-value pairs and `storage["key"] = value` assign them to the generic storage instance.
This can be wrapped in a try-except clause, so that when the assignment fails we `(cloud?)pickle` the object and try again, which would circumvent the current `h5io` failure for, e.g., ASE `Calculator` objects.
Replacing `_restore` with something based on `__setstate__` is a little more opaque to me, because I'm still not totally clear how `h5io` is getting me my initial classes, but maybe it will make more sense when I go back and look there.

Maybe we could also work in something extra like before- and after-storage hooks, e.g. for storing extra-state data like version controlling, so we can fail hard if we're deserializing the wrong version or whatever.
But that's a bell-and-whistle.

liamhuber:
Pre-`flowrep` issue

Issue #169, opened 2024-01-21
Hashing input to avoid running


liamhuber:
Nodes should optionally hash their input on run calls, and if they have this option enabled and find a "last hash" value, compare the current input to that hash and simply return the same output instead of running. (It might be good/necessary to hash the output too).

This will let us re-"run" graphs quickly in conditions where the hashing is much faster than the actual node computation. The hashing still gives us a speed-limit, but it seems like a good improvement.

Combined with saving and loading in #160, this will let us build more expensive workflows, run them, save them, then come back another day to extend them without having to re-run anything expensive we did earlier. 

When we introduce the ability to ship nodes off to a computation queue (e.g. slurm), if we do it in a way that the remotely executing job is serializing stuff to the right storage location, then combined with automatic loading in #160, this handles a big chunk of the work for fetching completed queue jobs too.

jan-janssen:
In `pyiron_base` we use the following hash function to hash the binary of the cloud pickled input plus function body https://github.com/pyiron/pyiron_base/pull/1285 :
```python
def get_hash(binary):
    # Remove specification of jupyter kernel from hash to be deterministic
    binary_no_ipykernel = re.sub(b"(?<=/ipykernel_)(.*)(?=/)", b"", binary)
    return str(hashlib.md5(binary_no_ipykernel).hexdigest())
```

The regex part is important to remove the jupyter notebook kernel which changes every time the jupyter notebook kernel is restarted.

liamhuber:
@jan-janssen, I really liked your comment in the meeting this morning that the hashing here should be kept distinct from hashing for the database; at the end of the day we may be able to reuse some functionality for actually generating the hashes, but I totally agree this should be an independent feature and can be developed separately from #126.

This issue is now very close to the top of my todo-list, and (mostly notes for myself here) the spec I have in mind is:
- Opt-in at the node instance level with a property (+ability to set via input kwarg)
- An easy method to clear an existing hash to force re-running (e.g. for when you've changed what the node functionality is)
- Allow the behaviour for anything where the input values hash natively, but just allow it to fail cleanly if you try to do it on a node with unhashable input -- this is the user's problem
- Maybe? Allow specifying the hashing property default at the class-level instead of directly in `__init__`, so that certain sub-classes wind up having opt-out instead of opt-in behaviour?
- Implementation question: can this be handled already in `Runnable` instead of down in `Node`?

PR #189, opened 2024-02-01
Add and test wrappers for sticking nodes in a pyiron job

liamhuber:
In the long run we want to have queue submission fully integrated with nodes; in the short term, we can get that by making a `pyiron_base` job out of a node. This isn't a permanent solution as it doesn't allow, e.g., sending just one node from a graph off to the queue, only entire graphs (even if they're just one node themself).

Here I introduce two solutions: one defines a new subclass of `TemplateJob` and relies on the node's own storage capabilities to (de)serialize itself (from)to HDF _inside_ the job's working directory. The only true `input` on the job is a string saying what storage backend the node should use (`"h5io"`/`"tinybase"`, per #160); then the user supplies a node instance to the job's `.node` attribute. The executed node can be found in the same place on the job on reloading.

The other exploits the [new](https://github.com/pyiron/pyiron_base/pull/1285) `pyiron_base.Project.wrap_python_function`, which takes a node in `job.input["node"]` and returns the executed version in `job.output["result"]` using the wrapper's `cloudpickle` and ignoring the nodes own storage.

Here's an example of the first approach copied from the docstring:

```python
        >>> from pyiron_base import Project
        >>> from pyiron_workflow import Workflow
        >>> import pyiron_workflow.job  # To get the job registered in JOB_CLASS_DICT
        >>> 
        >>> @Workflow.wrap_as.single_value_node("t")
        ... def Sleep(t):
        ...     from time import sleep
        ...     sleep(t)
        ...     return t
        >>> 
        >>> wf = Workflow("pyiron_node", overwrite_save=True)
        >>> wf.sleep = Sleep(0)
        >>> wf.out = wf.create.standard.UserInput(wf.sleep)
        >>> 
        >>> pr = Project("test")
        >>> 
        >>> nj = pr.create.job.NodeJob("my_node")
        >>> nj.node = wf
        >>> nj.run()
        >>> print(nj.node.outputs.to_value_dict())
        {'out__user_input': 0}

        >>> lj = pr.load(nj.job_name)
        >>> print(nj.node.outputs.to_value_dict())
        {'out__user_input': 0}

        >>> pr.remove_jobs(recursive=True, silently=True)
        >>> pr.remove(enable=True)
```

The docstrings and tests are relatively thorough, but I'm not keen to document it in the tutorial notebooks or expose the functionality directly on the `Workflow` class yet -- I'd rather if some people at MPIE played around with it a bit and let me know what would be helpful, which interface they prefer, etc.

The one big constraint in this PR is that the nodes will still only run on a single core, even when shipped off to a queue. This is because the node's `.executor` attribute is (so far) `None` or an instance of an executor -- and the later can't be easily serialized. @jan-janssen is working on this [for the underlying `PythonFunctionContainerJob`](https://github.com/pyiron/pyiron_base/pull/1296/files), but we may also be able to solve it independently for the `NodeJob` here using the same attack -- namely, supplying information on how to instantiate a fresh executor on the far side instead of passing along a live instance. We already have the infrastructure in place to do this in the [`Node._parse_executor` method](https://github.com/pyiron/pyiron_workflow/blob/f7e810c86a049f555836b41ddfc8d33878b6b15e/pyiron_workflow/node.py#L534), it's simply a matter of populating that method with the logic to parse a class name and args/kwargs for the new executor.

@ligerzero-ai, IIRC you were going to look at running multi-core Vasp jobs in a workflow? This should let you do it on the queue, so basically you just need to write parser nodes, and a shell node (preferably a generic one, and a more specific one that executes VASP in particular), then put them all together in a macro. @JNmpi has examples of these things for other codes over on his branch (#33), so we're not too far off.  I don't expect you to do anything with this immediately, I just want to stay mutually well informed so we don't wind up doubling up on work. If you do want to work on it, maybe you could first/also open a PR to extend `_parse_executors`?

**Note** that this PR depends on [pyiron_contrib](https://github.com/pyiron/pyiron_contrib/pull/949) and [h5io](https://github.com/h5io/h5io/pull/77) code that isn't merged to main yet, much less released, so none of it will run unless you clone those branches locally. This is also why the CI tests all fail.

### DIFF
--- a/pyiron_workflow/job.py
+++ b/pyiron_workflow/job.py
@@ -61,27 +61,21 @@ class NodeJob(TemplateJob):
         >>> from pyiron_workflow import Workflow
         >>> import pyiron_workflow.job  # To get the job registered in JOB_CLASS_DICT
         >>> 
-        >>> @Workflow.wrap_as.single_value_node("t")
-        ... def Sleep(t):
-        ...     from time import sleep
-        ...     sleep(t)
-        ...     return t
-        >>> 
         >>> wf = Workflow("pyiron_node", overwrite_save=True)
-        >>> wf.sleep = Sleep(0)
-        >>> wf.out = wf.create.standard.UserInput(wf.sleep)
+        >>> wf.answer = Workflow.create.standard.UserInput(42)  # Or your nodes
         >>> 
         >>> pr = Project("test")
         >>> 
         >>> nj = pr.create.job.NodeJob("my_node")
         >>> nj.node = wf
-        >>> nj.run()
+        >>> nj.run()  # doctest:+ELLIPSIS
+        The job my_node was saved and received the ID: ...
         >>> print(nj.node.outputs.to_value_dict())
-        {'out__user_input': 0}
+        {'answer__user_input': 42}
 
         >>> lj = pr.load(nj.job_name)
         >>> print(nj.node.outputs.to_value_dict())
-        {'out__user_input': 0}
+        {'answer__user_input': 42}
 
         >>> pr.remove_jobs(recursive=True, silently=True)
         >>> pr.remove(enable=True)
@@ -183,26 +177,19 @@ def create_job_with_python_wrapper(project, node):
         >>> from pyiron_workflow import Workflow
         >>> from pyiron_workflow.job import create_job_with_python_wrapper
         >>> 
-        >>> @Workflow.wrap_as.single_value_node("t")
-        ... def Sleep(t):
-        ...     from time import sleep
-        ...     sleep(t)
-        ...     return t
-        >>> 
         >>> wf = Workflow("pyiron_node", overwrite_save=True)
-        >>> wf.sleep = Sleep(0)
-        >>> wf.out = wf.create.standard.UserInput(wf.sleep)
+        >>> wf.answer = Workflow.create.standard.UserInput(42)  # Or your nodes
         >>> 
         >>> pr = Project("test")
         >>> 
         >>> nj = create_job_with_python_wrapper(pr, wf)
         >>> nj.run()
         >>> print(nj.output["result"].outputs.to_value_dict())
-        {'out__user_input': 0}
+        {'answer__user_input': 42}
 
         >>> lj = pr.load(nj.job_name)
         >>> print(nj.output["result"].outputs.to_value_dict())
-        {'out__user_input': 0}
+        {'answer__user_input': 42}
 
         >>> pr.remove_jobs(recursive=True, silently=True)
         >>> pr.remove(enable=True)
--- a/pyiron_workflow/job.py
+++ b/pyiron_workflow/job.py
@@ -42,7 +42,8 @@ _WARNINGS_STRING = """
 
 
 class NodeJob(TemplateJob):
-    __doc__ = """
+    __doc__ = (
+        """
     This job is an intermediate feature for accessing pyiron's queue submission
     infrastructure for nodes (function nodes, macros, or entire workflows).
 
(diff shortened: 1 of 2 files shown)


ligerzero-ai:
Cheers, I'll take a careful look over the next week. If it's urgent (stopping other work), feel free to merge w/o review, we can always revert :)

liamhuber:
> Cheers, I'll take a careful look over the next week. If it's urgent (stopping other work), feel free to merge w/o review, we can always revert :)

Super! No, this is pretty standalone so I should be fine leaving it open for O(week).

pyiron_workflow/job.py
```diff
@@ -0,0 +1,207 @@
...
+        if self._node is not None:
+            raise ValueError("Node already set, make a new job")
+        elif self._node_working_directory_already_there(new_node):
+            self.raise_working_directory_error()
+        else:
+            self._node = new_node
+            self.input._class_type = (
+                f"{new_node.__class__.__module__}." f"{new_node.__class__.__name__}"
+            )
+            self.input._label = new_node.label
```
ligerzero-ai, line 114: [not a blocker for this PR]

my main concern/feedback is that there are many cases where a user may not necessarily want to create a new folder for every single step of their workflow. Am I correct in thinking that this will raise an error if this reuses a working dir from a node? In a lot of cases, a workflow will fail at some point, and then will it not be extremely clunky to reuse early-computations from the earlier successful workflow steps. 

This is mostly a philosophical thing, and need not be addressed in this PR.

ligerzero-ai:
Some minor feedback - but otherwise lgtm. Let's get this merged :)

pyiron_workflow/job.py
liamhuber: > a user may not necessarily want to create a new folder for every single step of their workflow.

In this case each `NodeJob` is related 1:1 with a `Node`; that may be a single `Function` node, or a whole complex `Workflow`. In this respect, running a `NodeJob` is guaranteed to make at least one directory (for the job) and populate at least one file (the node's HDF file). So far we're no different from a regular pyiron job in this respect. If the `Node` is a graph and some of its children produce files, they'll be nested inside the job's directory with a directory structure that mirrors the graph topology; this is normal for the `Composite` nodes.

> Am I correct in thinking that this will raise an error if this reuses a working dir from a node?

What it's doing is making sure that the node is both fresh (not reloaded) _and_ that it hasn't peeked at its working directory -- which would set it to `./{node_label}`, when we want to force it to be nested inside the job `./{job_name}/{node_label}`. So a workflow that has been saved previously will surely fail, some nodes may be able to be assigned post-execution, but not if they already peaked at their nodes, e.g. if it's some macro running Lammps and the Lammps node has written IO files to the node's directory.

> In a lot of cases, a workflow will fail at some point, and then will it not be extremely clunky to reuse early-computations from the earlier successful workflow steps.
>
> This is mostly a philosophical thing, and need not be addressed in this PR.

These last two I'll take together. Yes. Yep, definitely still clunky. In general the current implementation of file storage is not yet  prepared for restarting from failed nodes -- it's a run-and-done feature right now. I agree the issue shouldn't be blocking as right now I'd prioritize getting some version of this capability running to build familiarity with the pain points, but I am fully on board with making sure we get philosophical regularly as we go along!

Ultimately I see wrapping the nodes in a job as an intermediate step for database and queue-access functionality. Ideally, I'd like to see each node be able to make these connections individually, e.g. so you could have some `Workflow` where _only_ the `Vasp` node got shipped off to the queue.

This would involve more sophisticated treatment of the storage, e.g. having remotely run nodes serialize to their own HDF5 file, then merging this into the main root-most node's HDF5 file at reload time. Some more thoughts on this are [here](https://github.com/pyiron/pyiron_workflow/issues/182).

Similarly instead of re-running a node, a key feature for recovering partially executed graphs is to re-load instead of re-run output -- i.e. we should be checking if there is saved data, if the current input matches the saved input, and if so just reload the old output instead of re-running. Discussed [here](https://github.com/pyiron/pyiron_workflow/issues/169) but also not implemented yet.

PR #191, opened 2024-02-01
Only subclasses

### LINKED ISSUE #190: Except for workflow, node behaviour should always be a class property
Right now we offer three fundamental ways of constructing a new node (either a `Function` or a `Macro`):
1. By decorating a function (which becomes the `Function.node_function` or `Macro.graph_creator`)
2. By subclassing `Function` or `Macro` and defining `Function.node_function ` or `Macro.graph_creator`
3. By instantiating a `Function` or `Macro` and setting `Function.node_function ` or `Macro.graph_creator` _as an instance property_.

So for the first two, information about how the node works (e.g. inputs and their names, and (at a minimum) how many outputs are returned) are available _at the class level_. While for the third route, by definition these are only available post-instantiation.

I would like to get rid of route (3) entirely and make `Function` and `Macro` abstract, so that actual instantiated nodes _always_ come from a class created either by a decorator or a manual sub-class.

Concretely, I want to disallow:

```python

def my_func(x):
    return x + 1

node = Workflow.create.Function(my_func)
```

In favour of the other two existing routes:

```python

def MyFunc(x):
    return x + 1

node = MyFunc()
```

or 

```python
from pyiron_workflow.function import Function

class MyFunc(Function):
    @staticmethod
    def node_function(x):
        return x + 1

node = MyFunc()
```

This will help get rid of some edge cases, [e.g. in storage](https://github.com/pyiron/pyiron_workflow/blob/edac929d644a7de283e47348127d4e61c7417138/pyiron_workflow/node.py#L178) where `h5io` has trouble reading instances of the base classes where these methods were provided at instantiation. 

It also streamlines other ideas, e.g. it will allow us to get type hints and scrape output labels at the class level; while this isn't critical now, being able to scrape info like type hints before we've instantiated things is exactly the type of tool we need for building node recommendation lists like in ironflow.

Bottom line: I think this removes a _little bit_ of flexibility for users, but gives us robustness and fewer edge cases to worry about. In the long run, I think it makes the platform stronger.

@samwaseda and @JNmpi, I'll push a PR on this and once it's ready I'll probably ping you (Sam), but it's more important to me that you give feedback on the idea than the execution.

liamhuber:
Closes #190

- [x] Make `Function`, `SingleValue`, and `Macro` abstract
- [x] Update docstrings
- [x] Update tests
- [ ] Update README
- [ ] Update notebooks

### DIFF
--- a/pyiron_workflow/function.py
+++ b/pyiron_workflow/function.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+from abc import ABC, abstractmethod
 import inspect
 import warnings
 from functools import partialmethod
@@ -16,7 +17,7 @@ if TYPE_CHECKING:
     from pyiron_workflow.composite import Composite
 
 
-class Function(Node):
+class Function(Node, ABC):
     """
     Function nodes wrap an arbitrary python function.
 
@@ -323,7 +324,6 @@ class Function(Node):
 
     def __init__(
         self,
-        node_function: callable,
         *args,
         label: Optional[str] = None,
         parent: Optional[Composite] = None,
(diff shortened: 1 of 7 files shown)


liamhuber:
Per the discussion in #190, let's better do this by modifying `__new__`

PR #199, opened 2024-02-07
Add a convenience method for reporting importability

liamhuber:
As discussed in [the parent branch](https://github.com/pyiron/pyiron_workflow/pull/197/files#r1481891075).

Actually, I super like this after all.

Screen Shot 2024-02-07 at 10 35 53

### DIFF
--- a/pyiron_workflow/composite.py
+++ b/pyiron_workflow/composite.py
@@ -768,8 +768,8 @@ class Composite(Node, ABC):
     def import_ready(self) -> bool:
         return super().import_ready and all(node.import_ready for node in self)
 
-    def import_readiness_report(self, tabs=0, report_so_far=""):
-        report = super().import_readiness_report(tabs=tabs, report_so_far=report_so_far)
+    def _report_import_readiness(self, tabs=0, report_so_far=""):
+        report = super()._report_import_readiness(tabs=tabs, report_so_far=report_so_far)
         for node in self:
-            report = node.import_readiness_report(tabs=tabs + 1, report_so_far=report)
+            report = node._report_import_readiness(tabs=tabs + 1, report_so_far=report)
         return report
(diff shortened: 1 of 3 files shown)


niklassiemer:
After the review of the other PR, I thought of the function to provide a dict of the report instead


```
    def import_readiness_report(self):
             return {self.label: self.import_ready}
         
```

```
    def import_readiness_report(self): 
         result=super().import_readiness_report(tabs=tabs) 
        if not result[self.label]:
             result[self label]={node.label: node.import_readiness_report() for node in self}
    return result
```

This would omit composites which are fine from being reported and one could provide this dict in the error upon save. Of course one can also make these private and add a nice printing user function. 

(sorry for missing formatting - ok my mobile 😅)

liamhuber:
> Lgtm, however tests seem to be violated?!

Did they fail for you locally? On my machine they run fine.

CI tests are fully expected to be broken because (a) I haven't updated my requested `pyiron_contrib` version yet, and (b) I have requested a version of `h5io` that isn't released yet 😝

liamhuber:
> After the review of the other PR, I thought of the function to provide a dict of the report instead
> 
> ```
>     def import_readiness_report(self):
>              return {self.label: self.import_ready}
>          
> ```
> 
>  ```
>     def import_readiness_report(self): 
>          result=super().import_readiness_report(tabs=tabs) 
>         if not result[self.label]:
>              result[self label]={node.label: node.import_readiness_report() for node in self}
>     return result
> ```
> 
> This would omit composites which are fine from being reported and one could provide this dict in the error upon save. Of course one can also make these private and add a nice printing user function.
> 
> (sorry for missing formatting - ok my mobile 😅)

I would actually prefer if the composite does register as failing.

Pros:
- Because trying to save it while any children are not saveable must crash
- IMO it makes it easier to trace down the report to where the broken leaf node is

Cons:
- The user might think they need to fix something in the composite defintion itself, even when that's fine.

liamhuber:
Now the readiness report appears right in the error message (good call, @niklassiemer!!) and `.import_readiness_report` is a property (for symmetry with `.readiness_report` and usability).

niklassiemer:
Like it :+1:
No time to actually play with the branch right now, only procrastinating before cleaning the kitchen 🤣

PR #213, opened 2024-02-19
Allow return None

samwaseda:
Currently the following function fails:

```python
@single_value_node()
def my_test(a=0):
    return None

test = my_test()
test
```

I'm actually not sure if we should make a distinction between returning `None` and not having `return` in the first place, but anyway the current error cannot be understood, so as long as we don't have a more explicit answer, I think we should allow any functions returning `None`.

### DIFF
--- a/pyiron_workflow/function.py
+++ b/pyiron_workflow/function.py
@@ -398,7 +398,7 @@ class Function(Node):
         dot-accessible.
         """
         parsed_outputs = ParseOutput(self.node_function).output
-        return [] if parsed_outputs is None else parsed_outputs
+        return [None] if parsed_outputs is None else parsed_outputs
 
     @property
     def _input_args(self):


liamhuber:
Nice catch.

> I'm actually not sure if we should make a distinction between returning `None` and not having `return` in the first place

I feel we definitely want to raise an error if the user tries to define a `Function` node (or child) without a `return` statement; that's clearly an error and should be caught as early as possible and with a decent message. Is it possible to distinguish in this way?

Otherwise, if there are multiple return values and one is `None` this works already. If there is a _single_ return value and it is _always_ `None`...that's weird? I mean, I guess I am ok with allowing it -- or at least, as you point out, failing with a much more informative error message -- but it's a pretty niche case.

Regardless of the behaviour we settle on, definitely missing from this PR is a test encoding the behaviour.

samwaseda:
> I feel we definitely want to raise an error if the user tries to define a `Function` node (or child) without a `return` statement; that's clearly an error and should be caught as early as possible and with a decent message. Is it possible to distinguish in this way?

I agree that we **want** to, but I don't really see why we **have** to, because for me raising an error is equivalent to "Sorry we cannot continue this way" and not "Look, I think what you did is not what you want". So even though I cannot imagine why someone should need a node in a workflow which returns nothing, I have the feeling we cannot forbid what python allows.

samwaseda:
> Regardless of the behaviour we settle on, definitely missing from this PR is a test encoding the behaviour.

Including a test which makes sure there's no error implies for me that `return None` must not raise an error, which is also not necessarily what I'm trying to achieve, because, from my perspective, it is not a problem that `return None` raises an error, as long as the error tells what went wrong and why it's an error. So I might be inclined to keep it open until we find a definitive answer to it.

samwaseda:
So I'm gonna merge this one. If you have a strong feeling about a test, feel free to add one.

liamhuber:
@samwaseda, the tests don't pass...

samwaseda:
XD I didn’t even realize

PR #214, opened 2024-02-19
File copy to

samwaseda:
There are cases where it makes more sense to copy files, such as LAMMPS potentials or pseudo potentials. In conventional pyiron it is not fully conceptualised, but I would like `FileObject` to be able to do it, so I added it here. I think the functionality is clear enough, but here's a short example:

```python
from pyiron_workflow.snippets.files import FileObject, DirectoryObject
my_directory = DirectoryObject(".")
my_file = FileObject(file_name="test.dat", directory=my_directory)
my_file.write("hello")
my_new_file = my_file.copy("new_test.dat")
print(my_new_file.read())
```

Output: `hello`


pyiron_workflow/snippets/files.py
```diff
@@ -145,3 +145,33 @@ def _resolve_directory_and_path(
         if not isinstance(directory, DirectoryObject):
             directory = DirectoryObject(directory)
         return file_name, directory
+
+    def copy(
+        self,
+        new_file_name: str | None=None,
+        directory: DirectoryObject | str | None=None
```
liamhuber, line 152: Use `Optional` and I guess the `directory` hint needs to be expanded to include `Path` now?

pyiron_workflow/snippets/files.py
```diff
@@ -145,3 +145,33 @@ def _resolve_directory_and_path(
...
+            new_file_name (str): New file name. You can also set
+                an absolute path (in which case `directory` will be ignored)
+            directory (DirectoryObject): Directory. If None, the same
+                directory is used
+        Returns:
+            (FileObject): file object of the new file
+        """
+        if new_file_name is None:
+            if directory is None:
+                raise ValueError(
```
liamhuber, line 166: You could consider introducing a new error type, like 

```python
class NoDestinationError(ValueError):
    """A custom error for when neither a new file name nor new location are provided"""
```

This is a habit I'm getting into as it makes it easier in the tests to make sure that I'm actually hitting the logic branch I'm expecting to hit.

liamhuber (approved):
I like it.

pyiron_workflow/snippets/files.py
samwaseda: No currently it doesn't include `Path`. It's a huge mess, I agree XD

liamhuber: > According to [what I see in the Internet](https://stackoverflow.com/questions/69440494/python-3-10-optionaltype-or-type-none), type hinting with `|` is preferred over `Optional`.

tldr; ok, then do whatever you want.

I'm actually quite ambivalent about this. On the one hand, I appreciate not needing to import `Optional`, on the other hand I feel that `foo: Optional[int] = None` is preferable because (a) it doesn't have `None` repeated, and (b) my brain can register that this is, in fact, optional only two characters after the `:`. In contrast, `foo: int | None = None` has ugly repetition and I can't use early stopping in my reading because I don't find out it's optional until the end. This is why ["There should be one-- and preferably only one --obvious way to do it."](https://peps.python.org/pep-0020/) 😠 😂

samwaseda: I agree that `Optional[str] = None` looks more elegant than `str | None = None`, but since there's too much work to do to care about elegance I'm gonna merge this one XD

samwaseda:
Ah wait I'm gonna merge it only when the target branches are merged, to make sure that none of the PR becomes too big

PR #216, opened 2024-02-20
Make DirectoryObject recognize various init variables

samwaseda:
I think it's straightforward enough, but this PR is to make `DirectoryObject` recognise `str`, `Path` and `DirectoryObject` for the directory instantiation.

I guess this is the first one to be merged. Sorry for going back and forth all the time!

### DIFF
--- a/tests/unit/snippets/test_files.py
+++ b/tests/unit/snippets/test_files.py
@@ -10,6 +10,12 @@ class TestFiles(unittest.TestCase):
     def tearDown(cls):
         cls.directory.delete()
 
+    def test_directory_instantiation(self):
+        directory = DirectoryObject(Path("test"))
+        self.assertEqual(directory.path, self.directory.path)
+        directory = DirectoryObject(self.directory)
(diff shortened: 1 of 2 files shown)


samwaseda:
@liamhuber I made [this project](https://github.com/orgs/pyiron/projects/13/views/1) to make a better overview. Hopefully it helps a bit.

tests/unit/snippets/test_files.py
```diff
@@ -10,6 +10,12 @@ def setUp(cls):
     def tearDown(cls):
         cls.directory.delete()
 
+    def test_directory_instantiation(self):
```
liamhuber, line 13: Missing the case of instantiation by string

liamhuber (changes requested):
Missing a test case, but otherwise lgtm

samwaseda:
Since the distinction won't be needed if [`DirectoryObject` is derived from `Path`](https://github.com/pyiron/pyiron_workflow/issues/218), I close this PR.

PR #220, opened 2024-02-20
pyiron_base's wrapper doesn't use storage

liamhuber:
It serializes the entire node to bytestream directly, so we should be able to run this job type on all supported python versions not just 3.11+.

### DIFF
--- a/pyiron_workflow/job.py
+++ b/pyiron_workflow/job.py
@@ -214,9 +214,6 @@ def create_job_with_python_wrapper(project, node):
     """
         + _WARNINGS_STRING
     )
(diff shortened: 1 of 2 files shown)


liamhuber:
```
OutputSignals ['ran'] (key node__index_0): DataContainer doesn't support saving elements of type "<class 'pyiron_workflow.workflow.Workflow'>" to HDF!
```

My guess is that under the hood we need `h5io`'s `__getstate__` for this to be working, so we're still restricted to >=3.11. IMO it's not worth digging deeper at this instant.

PR #225, opened 2024-02-23
[patch] NodeJobOutput: Overload the save function to not change the job name

jan-janssen:
@srmnitc This should solve the issue with the cryptic job names to be more consistent with `pyiron_base`.

PR #226, opened 2024-02-23
[patch] Exploit the mangling flag

liamhuber:
This depends on [`pyiron_base` 1356](https://github.com/pyiron/pyiron_base/pull/1356), and turns off the automatic name mangling for `NodeJob`. There is certainly not a large corpus of workflow jobs depending on the old naming scheme, so this can be a patch.

@jan-janssen @srmnitc, I don't know what the timetable is for the DPG or whether base releases can get made sufficiently quickly to utilize this, but it is far and away preferable to #225, which is wrongly patching a problem in `_base` but only inside `_workflow`.

### DIFF
--- a/pyiron_workflow/job.py
+++ b/pyiron_workflow/job.py
@@ -80,6 +80,9 @@ class NodeOutputJob(PythonFunctionContainerJob):
             raise NotImplementedError("Node jobs are only available in python 3.11+")
         super().__init__(project, job_name)
         self._function = _node_out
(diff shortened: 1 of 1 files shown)


liamhuber:
This won't be necessary, the next version of base will jump straight to [pyrion_base 1358](https://github.com/pyiron/pyiron_base/pull/1358/), so we'll get the desired renaming behaviour by default and never need to exploit the mangling flag

PR #227, opened 2024-02-23
[patch] Rely on the base class to not mangle names by default

liamhuber:
This reverts #226 in the event that [`pyiron_base` 1358](https://github.com/pyiron/pyiron_base/pull/1358) gets merged and the whole process is handled upstream.


liamhuber:
This won't be necessary, the next version of base will jump straight to [pyrion_base 1358](https://github.com/pyiron/pyiron_base/pull/1358/), so we'll get the desired renaming behaviour by default and never need #226

PR #233, opened 2024-02-28
Update draw

srmnitc:
This pull request updates the draw method to:

- add optional size option
- change colors
- turn off some gradient
- change font to sans serif helvetica


liamhuber (approved):
@srmnitc, stellar! I like the new look. I hijacked the PR to keep velocity up and added docstrings for the public facing parts that use the new kwarg, but that was my only nit. I also took the opportunity to propagate the change through to the educational material.

PR #234, opened 2024-02-28
[patch] Revert save overload

liamhuber:
Reverts #225. Waiting on #226 so we don't lose functionality of getting the requested name, which in turn is waiting on [`pyiron_base` 1356](https://github.com/pyiron/pyiron_base/pull/1356).


liamhuber:
With [pyiron_base 1358](https://github.com/pyiron/pyiron_base/pull/1358), the naming stays as requested by default when you actually instantiate a job from the class, so we need zero special behaviour in this repo to get the behaviour we want.

After the next base release, this can be re-tested on the updated version and merged.

PR #239, opened 2024-03-04
Synchronized dependency bumps

liamhuber:
So I don't need to think about who's dependency is whose or what is upper-limit locked, and since my bots are not updating the notebook env yaml anyway. Avoids running the whole CI suite N times.

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -7,9 +7,9 @@ dependencies:
 - cloudpickle =3.0.0
 - graphviz =8.1.0
 - h5io =0.2.2
-- h5io_browser =0.0.9
-- matplotlib =3.8.2
-- pyiron_base =0.7.8
+- h5io_browser =0.0.10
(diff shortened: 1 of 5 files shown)


liamhuber:
```
├─ pyiron_atomistics 0.4.17**  is installable and it requires
│  └─ pyiron_base >=0.7.7,<=0.7.9 , which can be installed;
└─ pyiron_base 0.7.10**  is not installable because it conflicts with any installable versions previously reported.
```

liamhuber:
```
├─ h5io_browser 0.0.10**  is requested and can be installed;
└─ pyiron_base 0.7.9**  is not installable because it requires
   └─ h5io_browser >=0.0.6,<=0.0.9 , which conflicts with any installable versions previously reported.
```

Well, I guess at least I will only run the _successful_ CI once...

liamhuber:
The rest waits on `pyiron_atomistics`

PR #240, opened 2024-03-04
JNmpi patch main head

liamhuber:


liamhuber:
`Node.iter` is not playing nicely with `Node.__post__`. As far as I can tell, `iter` is shipping things off to an executor which is triggering the `post` call multiple times for the same node; if I understanding this is giving a race condition for the different threads that results in an error if one of them manages to call the final `self.graph_root.tidy_working_directory()` line of `__post__` before they've all finished the earlier `self.storage.has_contents` check.

There might be some workaround, but I guess this is a fundamental incompatibility between the assumptions of the storage branch -- i.e., that each node is unique and has full sovereignty over its file and semantic path -- and `iter` -- which is re-instantiating new nodes with all the same kwargs and running them in parallel. Naturally the multiple copies fight with each other over the file path, since they all have the same name.

liamhuber:
I'll introduce compatible for-loop and a dataclass node before trying to merge these.

Issue #48, opened 2023-10-23
Running with an executor _sometimes_ hits a bug


liamhuber:
It's quite infrequent (< 1 in 10), but sometimes [this test](https://github.com/pyiron/pyiron_workflow/blob/696ff1488eed0c333e8700fe7aeca4a9ce80d09c/tests/unit/test_macro.py#L522) fails in the following way:

```
======================================================================
FAIL: test_with_executor (test_macro.TestMacro.test_with_executor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:\a\pyiron_workflow\pyiron_workflow\tests\unit\test_macro.py", line 522, in test_with_executor
    self.assertIs(
AssertionError: <pyiron_workflow.macro.Macro object at 0x0000021B611DB610> is not None : Returned nodes should get the macro as their parent

----------------------------------------------------------------------
```

I.e. when a composite is run on an executor and copies of its children are returned as the result, these new children are somehow failing to get the parent-child relationship set up reflexively (which is the expected behaviour). I vaguely suspect it has something to do with `Future.result()` returning new instances each call, or some sort of delayed execution that I am wrongly assuming is happening synchronously, or similar, but working with executors is pretty new to me and the error is not consistently reproducible, so I haven't managed to track it down yet.

liamhuber:
Happened again today. Was on the Windows 3.11 CI (I want to start noting this here in case there is some pattern of where it comes up, e.g. on windows)

liamhuber:
Happened again on the windows 3.11 CI

1/10 seems like an overestimate of how often it crops up, and so far really seems like a windows thing

liamhuber:
Happened on Ubuntu 3.11 today, so I guess it's not just windows.

liamhuber:
Happened on Ubuntu 3.11 today.

I can't remember the last time it happened on my local machine, so maybe it's exclusively an issue in the CI environment?

liamhuber:
Seems to be happening consistency today in #70, so maybe there's a chance to nail this down

liamhuber:
Ok, I added a little sleep and then suddenly the CI passed, so maybe this is fixed after all? Could it be simply that the callback function is executing to slowly on the CI (but fast enough on my local machine) that the test comes up before the state is ready for it?

```python
returned_nodes = result.result()  # Wait for the process to finish
        from time import sleep  # The "fixing" change (so far)
        sleep(1)  # The "fixing" change (so far)
[...]

PR #241, opened 2024-03-11
Add error message for return item without `channel`

samwaseda:
I keep making a mistake like this:

```python
@single_value_node("sum")
def addition(a, b):
    return a + b

@single_value_node("product")
def multiplication(a, b):
    return a * b

@macro_node("result")
def add_and_multiply(macro):
    macro.addition = addition(1, 2)
    macro.multiplication = multiplication(macro.addition, 5)
    return macro.multiplication.outputs.product.value

node = add_and_multiply()
```

The main problem is that in the current implementation it only says `NOT_DATA` does not have `channel`, which doesn't immediately make me understand where the error is coming from.

### DIFF
--- a/pyiron_workflow/macro.py
+++ b/pyiron_workflow/macro.py
@@ -317,13 +317,10 @@ class Macro(Composite):
         if len(ui_nodes) > 0:
             self._whitelist_inputs_map(*ui_nodes)
         if returned_has_channel_objects is not None:
+            if not isinstance(returned_has_channel_objects, tuple):
+                returned_has_channel_objects = (returned_has_channel_objects,)
             self._whitelist_outputs_map(
-                output_labels,
-                *(
-                    (returned_has_channel_objects,)
-                    if not isinstance(returned_has_channel_objects, tuple)
-                    else returned_has_channel_objects
-                ),
+                output_labels, *returned_has_channel_objects
             )
 
         self._inputs: Inputs = self._build_inputs()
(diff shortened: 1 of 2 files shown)


liamhuber (commented):
I totally agree with the diagnosis, and that the proper outcome is to raise a more informative error, I only have a differing opinion on finer details of the treatment. My disclaimer is that atm I'm on the couch on mobile with a sick kid napping on me 😂 so I can't browse the code base easily. I'll try to come back and polish this if I get a chance, but in case you get to it first I wanted you to have the feedback.

My main concern is that I don't like the location of the check -- IIRC this gets used for both I and O at some point, but this is strictly an O problem; it also un-statics the method, and while I agree we need self for an informative message (label and options), it is a yellow flag to me to convert away from static.

I think I would prefer a new private `_prepare_...` method analogous to the input one and to do this there. At any rate, some other more specific and more specifically named location for this work please!

Otherwise just two nits:
- I'm trying to reduce string usage in the messages to help refactoring be more stable; this might be an edge case because of what we want to say, but you could play around with sentence formulations that replace the `"channel"` string with `{HasChannel.__name__}`
- I'm not sold on giving a link to the repo in the error message. IMO people know what the package name is and will find their way there by themselves. If we _do_ wdecide to include this, I would want to do so consistently across all user-facing messages (and thus in a separate PR)

liamhuber:
Oh, and we need a test added that catches the new message

liamhuber:
Ok, I've got my computer and I stand by everything I said earlier. More specifically, somewhere in here:

https://github.com/pyiron/pyiron_workflow/blob/bb326ec6d26efd371eac45d431c0d3d6d1dc680a/pyiron_workflow/macro.py#L313-L319

I'd slot in a new method like `self._ensure_returned_objects_are_has_channels` (or similar) that does the check you want. This is perfectly in line with the variable name, it's just that right now the code _assumes_ that the user has correctly returned only `HasChannel` objects. I totally agree that's an easy mistake though, so let's add the filtering. To fail as early as possible I recommend inserting the check immediately after line 313. 

In the future we could even imagine extending the check to create channel objects from non-HasChannel returns. This heads in the direction of getting macro and function nodes to be more similar, and is absolutely not necessary here and now, but I like that this change is commensurate with that direction.

samwaseda:
Ok I messed around. In the end I didn't put it in `__init__`, because for my taste there are already too many lines. I also added a test.

pyiron_workflow/macro.py
```diff
@@ -317,13 +317,10 @@ def __init__(
         if len(ui_nodes) > 0:
             self._whitelist_inputs_map(*ui_nodes)
         if returned_has_channel_objects is not None:
+            if not isinstance(returned_has_channel_objects, tuple):
```
samwaseda, line 320: This was not intentional (also nothing changed). I can revoke this change if you prefer.

liamhuber: Maybe we lack a test where the graph creator returns only a single value?

liamhuber: Ah, nvm, I see your comment applies to the change in the next few lines too. It's just a refactor. I don't have strong feelings, I find both perfectly readable and don't see an efficiency difference.

liamhuber (approved):
LGTM now! The windows failure is just #48. I want the tests passing before we merge, but I would be surprised if it takes more than one re-run as the error is exceptionally rare.

liamhuber:
Ugh, my rerun failed jobs button has no effect. I'm not sure if it's a problem with my browser or what, but this happens sometimes. @samwaseda could you rerun the failed windows test before merging?

Issue #243, opened 2024-03-13
Decompose `Node`


liamhuber:
The `Node` class has gotten a bit unwieldy. I feel that the way I added storage, especially on top of two backends that are non-optimal (`tinybase` being incomplete, and `h5io` -- most of all -- not accommodating rercursive attributes) has resulted in some technical debt. 

This became particularly evident in #240 when I tried to merge storage into Joerg's branch. I knew already that `Node.iter` didn't play well with the graph paradigm, but with storage present simply creating new copies of the same node with 

So what I'd like to do is break apart `Node` into some different pieces to separate out the concerns and break responsibilities into smaller pieces. I had initially tried a purely component-based composition approach (e.g. `Node.semantics.label` instead of `Node.label`), but this proliferated trouble with `h5io` due to all the recursive relationships. So instead I'm opting for mutliple-inheritance mixin-compositions.

My current attack plan is:
- [x] `Semantics` to handle labels, uniqueness, child/parent relationships, etc. (#244)
- [x] `Runnable` (#246)
  - And later to handle interactions with remote queues
- [x] `HasIO` to (#252)
  - [x] Use case: "single-value-ness" becomes a special case of `HasIO` where there is only one output data channel, so we can close #155 _and_ reduce the number of user-facing classes (just `Function` and `Macro`, no more `SingleValue`) at the same time. (#255, #257)
- [x] Finally, go back and revisit the existing abstraction of storage (#260)

liamhuber:
Ok, I'm quite happy with this. I'm quite certain that there is room for improvement, and it's always possible that one goes too far decomposing things into too small of pieces, but overall this feels better. IMO the proof is in the pudding with how easy it was to now give _all_ nodes "single value" power in #255. Critically, `pyiron_workflow.node` went from an unwieldy 1302 lines down to 729 -- a 46% reduction. There will be more to do in the future, but I feel this is a reasonable stopping point to go harvest some fruit from the improved infrastructure and get feature implementation done.

PR #244, opened 2024-03-13
[minor] Semantics

liamhuber:
Breaking out responsibility for the label, parent/child relationships (including semantic uniqueness), path, root, etc. into  separate classes. Long term, this is also useful for non-workflow/non-node semantic objects used to indicate non-data connections between workflows.

Part of #243 

Only user-facing changes are that `Composite.nodes` got renamed to `Composite.children` and is a `bidict` instead of `DotDict`.


liamhuber:
Test failures were a timeout downloading from conda, I guess this is 100% on the GitHub side and will just re-run.

PR #245, opened 2024-03-14
[patch] Refactor interfaces

liamhuber:
This introduces a new module `has_interface_mixins` to hold all the classes that are just there to guarantee the another class has some particular attribute or method. Such mixins are useful when we decompose responsibility into different components, but component A needs to know about some particular aspect of component B -- e.g., re:#243,  `HasIO` will need to know about the `run` method, but it won't care _at all_ what it does, so `Runnable` and `HasIO` will both inherit from `HasRun`, but for `HasIO` this is just to guarantee that the IO mixin gets combined with some _other_ component that defines `run`.

I intentionally ignore the `HasToDict` mixin here, because (a) it does somewhat more than simply guarantee the interface, and (b) it's on death row anyhow, as it's just there for prettier representations and I think we're going to be able to accomplish that more elegantly in other ways soon.

This is all under-the-hood, so no changes to the user interface.


liamhuber:
Well shoot, squashing and merging the semantic branch means all the squashed commits here still show up! I'm sure there is some sort of rebasing scheme that allows squashing to play well with stacking, but it's not something I want to work out on a Saturday, so for the rest of the stack I'm not going to squash.

PR #246, opened 2024-03-14
[minor] Extract responsibility for `run` to its own class

liamhuber:
This continues #243 by pulling out responsibility for handling run behaviour, that means having a status and talking to the future and handling callbacks. `Runnable` doesn't know anything about IO or graphs, so `Node` still implements all the different run modes (pulling, etc).

This is only intended to be a pure refactoring, so while I'm definitely interested in feedback on the architecture of `Runnable`, I might delay implementing changes that would echo into the downstream classes to a future PR.

~This is all under-the-hood, so no changes for users or node designers.~ Setting `check_readiness=False` in `run` will now run even if readiness failed due to a status issue; so you can really just brute-force run un-ready things with this flag.


liamhuber:
In the process I uncovered that the `while` meta-node is subject to recursion errors waaaay earlier than I'd like. I tested dropping the test threshold from 0.1 to 0.001 on `main` as well, and you hit the exact same recursion error, so this problem is quantitatively exacerbated by pulling out a superclass, but qualitatively unrelated to the changes in this PR; I'll open an issue and patch the meta-node's docstring with a heads-up separately.

PR #258, opened 2024-03-22
[patch] Has working directory

liamhuber:
Introduce an interface for having a parent and a mixin for having a working directory.

Contributes to #243


liamhuber:
Oops, renamed a branch and then pushed again instead of making a new branch and pushing. Fixed via force-push.

PR #265, opened 2024-03-26
[minor] Make `Function` and `Macro` definition functions available at the class level

### LINKED ISSUE #190: Except for workflow, node behaviour should always be a class property

liamhuber:
I'm still hacking away at getting Joerg's `iter` functionality and node library integrated into the codebase in a robust, storage-compatible way. Most of the attacks I've thought of involve knowing about the IO (labels and hints) at the class level, so I stopped and came back to #190, which wants to make `Function.node_function` and `Macro.graph_creator` both class methods.

The attack I'm taking is to make the existing `Macro` and `Function` calls actually be functions that dynamically create and return children of `AbstractMacro` and `AbstractFunction` (formerly the `Macro` and `Function` classes, respectively) by invoking the decorator. This approach made the refactor for `Function` super duper easy and I'm quite happy with it so far.

TODO:
- [x] Repeat the attack for `Macro`
- [x] Update docs ~and notebooks~

Closes #190 

Non-goals: 
- Actually doing anything with this class method (like exploiting it to expose default IO channel patterns or similar) -- this is a fine idea but a separate PR.
- Discussing this in the tutorial notebooks -- we want people using the decorators anyhow

Minor bump because any nodes inheriting from `Function`/`Macro` now need to inherit from `AbstractFunction`/`AbstractMacro` (afaik there are no such nodes outside this repo, but better to be safe).


liamhuber:
@pmrv I thought you might find this somewhat interesting given our discussion in #190. I wound up leveraging `__new__` after all, but instead of doing it in the real inheritance tree, it's happening in a little shell class that doesn't do much other than dynamically create the new class! Basically these are just class-like alternatives to the decorator functions!

pmrv:
I'm not sure I follow all the call sites, but if `func = Function(...)` just makes a new dynamic class and `not isinstance(func, Function)` holds, wouldn't it be simpler and less indirect to just have `Function` as a plain function or method on the creator where it will end up being called from?

liamhuber:
@pmrv, yes, honestly this was probably an instance of "Your Scientists Were So Preoccupied With Whether Or Not They Could, They Didn't Stop To Think If They Should". There is some rationale to keep the callable PascalCase, since in this frameworks that indicates to our users that they're getting a `Node` instance back, and a big block of our target audience is people who aren't even aware of OOP. So making it a class has the advantage that my IDE doesn't complain to me that I'm casing a function wrong 😂 But you're absolutely right, while there's no operational difference between a function returning the new instance and a class that _only_ overrides `__new__` and returns the same new instance, the direct function is easier to maintain.

I'll try switching it back to a function later today. I don't anticipate any complications, but if any arise I'll bail on this and just delay it until the planned refactoring where I extract a bunch of shared behaviour between `AbstractFunction` and `AbstractMacro` into a mutual parent.

liamhuber:
@pmrv right, ok, I remember now why I did it this way. For `Function` it absolutely could just be an actual function and not a class. But for `Macro`, `AbstractMacro` (and `Workflow`, and everything that inherits from `Composite`) has access to the creator at the class-level. Since `Function` and `Macro` _look_ like node classes, I wanted to be able to offer the same creator access for consistency. E.g. one could, in principle, write

```python
from pyiron_workflow.macro import Macro

def whatever(macro, a, b):
    macro.add = Macro.create.standard.Add(a, b)
    return macro.add

m = Macro(whatever, output_labels="sum")
```

I'm still open to renaming `Function`->`function` and `Macro`->`macro` so they look like functions again and then get rid of this feature, but because of all the places `Function` and `Macro` get called that will almost certainly wind up causing inane merge conflicts as I try to pull the change up through the PRs stacked on top of this one...so if we're going to do it, let's do it later.

PR #266, opened 2024-03-27
[minor] Make `Function` IO info available at the class level

liamhuber:
This will be useful for building nodes that take other node classes as input and want to build their own corresponding IO based on that class's IO without having to instantiate it first. It also borders on necessary for the long-term goal of being able to suggest nodes to users based on ontological (or regular...) type hints _ala_ `ironflow`.

Example:
```python

@Workflow.wrap_as.function_node("xplus1", "xminus1")
def PlusMinusBound0(x: int) -> tuple[int, int | None]:
    return x + 1, None if x - 1 < 0 else x - 1

print(PlusMinusBound0.preview_output_channels())
>>> {'xplus1': int, 'xminus1': int | None}

print(PlusMinusBound0.preview_input_channels())
>>> {'x': (<class 'int'>, NOT_DATA)}
```

Tasks:
- [x] Make output info available
- [x] Make input info available
- [x] Update docs/notebooks as needed (I don't see a need to update the notebooks -- until there is a better use case in, e.g. for loops or whatever, even node developers don't need to care about this functionality)

Non-goals:
- The same treatment for `Macro`, which makes sense but is left to another PR
- Anything like this for `Workflow`, which doesn't make sense since its IO is dynamic anyhow

Minor bump as the signature for `AbstractFunction` has changed.


liamhuber:
Windows unit tests failed with 

```
graphviz.backend.execute.CalledProcessError: Command '[WindowsPath('dot'), '-Kdot', '-Tpdf', '-O', 'start_graph']' returned non-zero exit status 1. [stderr: b'Warning: Could not load "C:\\Miniconda3\\envs\\my-env\\Library\\bin\\gvplugin_pango.dll" - It was found, so perhaps one of its dependents was not.  Try ldd.\r\nWarning: Could not load "C:\\Miniconda3\\envs\\my-env\\Library\\bin\\gvplugin_pango.dll" - It was found, so perhaps one of its dependents was not.  Try ldd.\r\nFormat: "pdf" not recognized. Use one of: bmp canon cmap cmapx cmapx_np dot dot_json emf emfplus eps fig gd gd2 gif gv imap imap_np ismap jpe jpeg jpg json json0 metafile mp pdf pic plain plain-ext png pov ps ps2 svg tif tiff tk webp xdot xdot1.2 xdot1.4 xdot_json\r\n']
```

Doesn't look directly related to these changes, but I might need to go back and pin graphviz to some other version.

liamhuber:
Yyyep.

```python
ERROR: test_draw (test_node.TestNode.test_draw)
  File "C:\Miniconda3\envs\my-env\Lib\site-packages\graphviz\backend\execute.py", line 88, in run_check
    proc.check_returncode()
  File "C:\Miniconda3\envs\my-env\Lib\subprocess.py", line 502, in check_returncode
    raise CalledProcessError(self.returncode, self.args, self.stdout,
subprocess.CalledProcessError: Command '[WindowsPath('dot'), '-Kdot', '-Tpdf', '-O', 'start_graph']' returned non-zero exit status 1.
```

liamhuber:
Windows error needs to be fixed, but it's from graphviz and unrelated to these changes.

PR #269, opened 2024-03-28
[minor] Make macro output labels a class attribute

liamhuber:
Minimal changes to remove `output_labels` from the signature of `AbstractMacro` and make these a class property (`_provided_output_labels`), just like in `AbstractFunction`.

Non-goals:
- Refactoring the shared functionality between `AbstractMacro` and `AbstractFunction` out into a mixin


liamhuber:
```
FAIL: test_draw (test_node.TestNode.test_draw) [Testing with format pdf]
  File "D:\a\pyiron_workflow\pyiron_workflow\tests\unit\test_node.py", line 326, in test_draw
    with self.assertRaises(
AssertionError: CalledProcessError not raised : Graphviz doesn't seem to be happy about the combindation PDF format and Windows right now. We throw a warning for it in `Node.draw`, so if this test ever fails and this combination _doesn't_ fail, remove this extra bit of testing and remove the warning.
```

Looks like upstream might have patched windows already...

PR #278, opened 2024-04-05
[minor] Be direct instead of using weird creator classes

liamhuber:
Per [@pmrv's comment](https://github.com/pyiron/pyiron_workflow/pull/265#issuecomment-2036454897), the `Function` and `Macro` classes that overrode `__new__` to return children of `AbstractFunction` and `AbstractMacro` we an unnecessary complication -- we can do the same thing with a plain old function. Here we transform those creators back into functions, and then give the base classes back their original names of `Function` and `Macro`.

@pmrv the main thing I want review on is to see if you have any better ideas for naming. Right now (class, decorator -> class, creator -> instance) are (`Function`, `as_function_node`, `function_node`) and (`Macro`, `as_macro_node`, `macro_node`). I also re-named `Creator.wrap_as` to `Creator.wrap`; so before/after goes like `Workflow.create.Function`/`Workflow.create.function_node`, `Workflow.wrap_as.function_node`/`Workflow.wrap.as_function_node`, etc..


pmrv (approved):
lgtm, the new naming works, at least I didn't think of any alternatives quickly.

PR #279, opened 2024-04-05
[minor] Function selflessly

liamhuber:
Currently, we jump through some hoops to let users pass `self` as a special argument for `Function` nodes. This didn't show up in the node input, and under the hood the node instance got passed in for that argument. You could do things like this:

```python

def Selfish(self, x):
    self.some_attribute = 5
    return x

n = Selfish()
print(n.inputs.labels)
>>> ["x"]

n(x=42)
print(n.some_attribute)
>>> 5
```

Now we don't, and `self` is treated just like any other `Function` node argument, which means it bumps right into the protection we have in place to stop people from using terms already in the `Function.__init__` signature (this protection allows input to be set as `**kwargs`)

```python

@Workflow.wrap.as_function_node()
def Selfish(self, x):
    self.some_attribute = 5
    return x

n = Selfish()
print(n.inputs.labels)
>>> ValueError: The Input channel name self is not valid. Please choose a name _not_ among ['self', 'args', 'label', 'parent', 'overwrite_save', 'run_after_init', 'storage_backend', 'save_after_run', 'kwargs']
```

I like for a bunch of reasons:
- Function nodes should behave functionally, so let's not explicitly give the user tools to break that
- The `self` argument didn't play well with executors, so there were special exceptions to fail early and non-universal behaviour
- Accommodating this took extra special-case code and increased maintenance cost
- It seems very in line with @JNmpi's sentiment at the last meeting about how nodes should not be having extra properties

I.e. if you want state, use a state-ful `Macro` node and hold the state as IO of one of its children.

Non-goals: 
- Failing early, like when the node class is defined, instead of waiting for the `input` to be accessed. I want this, and it should be easy, but it pertains to _all_ protected signature items and is out of scope here.

EDIT: python syntax highlighting on the examples

### DIFF
--- a/pyiron_workflow/function.py
+++ b/pyiron_workflow/function.py
@@ -428,19 +428,8 @@ class Function(Node, ABC):
         """
         type_hints = cls._type_hints()
         scraped: dict[str, tuple[Any, Any]] = {}
-        for ii, (label, value) in enumerate(cls._input_args().items()):
-            is_self = False
-            if label == "self":  # `self` is reserved for the node object
-                if ii == 0:
-                    is_self = True
-                else:
-                    warnings.warn(
-                        "`self` is used as an argument but not in the first"
-                        " position, so it is treated as a normal function"
-                        " argument. If it is to be treated as the node object,"
-                        " use it as a first argument"
-                    )
-            elif label in cls._init_keywords():
+        for label, param in cls._input_args().items():
+            if label in cls._init_keywords():
                 # We allow users to parse arbitrary kwargs as channel initialization
                 # So don't let them choose bad channel names
                 raise ValueError(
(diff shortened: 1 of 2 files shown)


samwaseda:
Sorry I had a heated discussion yesterday elsewhere and could not join the pyiron meeting. I was never very comfortable with `self` (I guess same for you) so I'm ok with getting rid of it. But this only removes `self` right? I thought the intention of `self` was to pass information from outside, like:

```python

@Workflow.wrap.as_function_node()
def Selfish(self, some_executable):
    some_executable.cores = self.server.cores
    some_executable.run()
    return some_executable.output
```

Or do we now want to do everything via function arguments?

PR #315, opened 2024-05-07
JNmpi lammps nodes patch version ~0.7.0~ 0.7.1

liamhuber:


liamhuber:
@JNmpi, these are the minimal changes (at least that's my attempt...) to get your node packages synced with the main branch. Here my goal was only to get the notebooks running, and I didn't purge any extraneous tools or check all the nodes. 

Now that the main branch has for-loop stuff (including an `.iter` method), the changes here are all inside `node_library`. My next step is to get a `pyiron_nodes` repo set up as discussed and migrate the non-`standard` nodes over there, including what is left in `dev_tools` for later cleanup/exploitation.

PR #329, opened 2024-05-14
[minor] Convert to pyproject.toml

liamhuber:
Including changes to identifiers and authorship, and new release CI, as well as a temporary CI script to test out the build. I'm omitting many build requirements as an experiment.

A minor bump insofaras the conda release will have different dependency limits


liamhuber:
The crux of the issue is that I'm a massive idiot and I didn't add

```python
from ._version import get_versions

__version__ = get_versions()["version"]
```

to `pyiron_workflow.__init__`. 

This looks like it's working though and I'll come back and clean it up (including git history) soon.

liamhuber:
I fought with this a lot, but ultimately the problem was extremely simple: I had forgotten to add `from ._version import get_versions; __version__ = get_versions()["version"]` to the module's init file. The struggle resulted in a few nice little QoL changes to pyiron/actions, but I wish I had found a quicker path. I reset to main and force-pushed over all my blundering to keep the history cleaner.

Issue #350, opened 2024-06-07
:bug: readthedocs build is failing


liamhuber:
readthedocs log:

```
        × Building wheel for mpi4py (pyproject.toml) did not run successfully.
        │ exit code: 1
        ╰─> [228 lines of output]

        ...

            --------------------------------------------------------------------------
            The Open MPI wrapper compiler was unable to find the specified compiler
            x86_64-conda-linux-gnu-cc in your PATH.
      
            Note that this compiler was either specified at configure time or in
            one of several possible environment variables.
            --------------------------------------------------------------------------
            failure.

        ...

```

The same issue happens for `pyiron_module_template`, which only has the dependency `pyiron_base`, so the problem is not with the actual source code. The `pyiron_base` docs themselves, however, are passing, so there must still be a repo-specific problem here. I suspect I can config my way out of this problem.

jan-janssen:
`pyiron_base` uses:
```
pip install versioneer[toml]==0.29
pip install . --no-deps --no-build-isolation
```
the `--no-build-isolation` part is important to prevent pip from building all dependencies.

liamhuber:
I am failing before the pip install

liamhuber:
<img width="916" alt="Screen Shot 2024-06-07 at 09 26 45" src="https://github.com/pyiron/pyiron_workflow/assets/20283329/621e89e2-1d7c-4c37-a3ef-0bd7620c8b3e">

liamhuber:
The versioneer[toml] and build dependency made it into the docs check as a part of the pyiron/actions upgrades, but I failed to propagate it to `.readthedocs.yml` throughout.

liamhuber:
It is still running `python -m pip install --upgrade --upgrade-strategy only-if-needed --no-cache-dir .` instead of `pip install versioneer[toml]==0.29; pip install . --no-deps --no-build-isolation`, my guess is that the trailing `python: install:` section of my yml is overriding the pre-build job.

PR #351, opened 2024-06-07
[patch] Update readthedocs config

liamhuber:
Bumping some versions, adding the versioneer[toml] install and build isolation, and removing other install directions


liamhuber:
Closes #350

PR #362, opened 2024-06-11
[minor] Move stuff

liamhuber:
This is a refactor, but since there was not a concrete API and things are in different places, it's a minor rather than patch bump.

I took everything downstream of `Node` (except `Workflow`) and moved it into the submodule `nodes`, and everything upstream of `Node` (or mixed into a subclass) and `Channel` and moved it into the submodule `mixin`. IMO this makes the entire project structure easier to understand.

While users are still intended to use `Workflow` as a single-point-of entry -- and supported in this by the creator/wrapper menus -- I also found that people developing nodes seem to like to import stuff like `as_function_node` _directly_ and use it. This is now explicitly supported by a "node developer API" section of the `__init__` file which has all the standard factories, decorators, creators, and parent classes that should be necessary for day-to-day creation of new node packages.


liamhuber:
I considered and (for now) rejected the idea of moving things the nodes/channels use but don't inherit from into their own submodule (e.g. `draw`, `parse_output`, `topology`, etc, into `tools` or similar). I found unlike the mixins/nodes which could be clearly identified by their up/downstream position in the `Node` inheritance hierarchy, for these objects there was no such clear-cut way to distinguish what should be moved into the submodule and what should be left at the main level.

It is still clear that something like `topology` is more niche than something like `io`, so I see room to further pare down the main level of the module in the future -- I just don't have a clear vision for how to do it right now.

PR #386, opened 2024-07-24
[minor] Switch from pympipool to executorlib

jan-janssen:


liamhuber:
This changes the default executor class, so it is not a patch change -> minor

liamhuber:
Draft because I think I'd like to merge the stack top-to-bottom here, so main gets the desired state in a single go.

Issue #387, opened 2024-07-24
Using macOS-11 for CI results in long wait times


jan-janssen:
**Summary**

Both `executorlib` and previously `pympipool` could already use `macos-latest` since the introduction of the `hostname_localhost` parameter at the beginning of the year: https://github.com/pyiron/executorlib/pull/258

**pyiron Version and Platform**

macOS 

**Expected Behavior**

By adding the `hostname_localhost` parameter for the CI, the CI wait time is reduced. 

**Actual Behavior**

Currently `macos-11` is required as later versions like `macos-14` do not allow connections to `localhost` on macOS. 

**Steps to Reproduce**

It happens at every pull request

**Further Information, Files, and Links**

-

liamhuber:
The only reason I used `macos-11` was to accommodate `pympipool`, so I am totally fine updating it to `macos-latest` -- it had just been so long since I looked at https://github.com/pyiron/executorlib/pull/258 that I forgot about it!

In #388 the `macos-latest` tests started basically right away; whereas the hang for `macos-11` without the localhost business was effectively infinite. From the issue it's a bit unclear -- is `macos-11`+localhost _even faster_ than `macos-latest`? I see you use latest in `executorlib`, so my guess is this is your recommended course of action here too?

jan-janssen:
> In #388 the `macos-latest` tests started basically right away; whereas the hang for `macos-11` without the localhost business was effectively infinite. From the issue it's a bit unclear -- is `macos-11`+localhost _even faster_ than `macos-latest`? I see you use latest in `executorlib`, so my guess is this is your recommended course of action here too?

No `macos-latest` requires `hostname_localhost` otherwise the tests fail.

liamhuber:
Ok, that's not a fundamental problem, but it does mean I want to change the default `Workflow.create.Executor` away from `executorlib.Executor`, since I'd need to add a bunch of `executorlib`-specific kwargs throughout the tests. I do this in #389, where we still use the `executorlib` one in notebooks (process pool fundamentally cannot handle that case since locally defined nodes become unimportable).

Since `ProcessPoolExecutor` doesn't work in notebooks, and notebooks are the key expected use case for the package, I'm now considering getting rid of the `Creator.Executor` entirely and making users explicitly choose one instead of giving a biased "default".

PR #388, opened 2024-07-24
[minor] Executorlib ci and docs

liamhuber:
A follow-up to #386 that purges references to `pympipool` in the docs and executor class name.

I also switch to `macos-latest` as `executorlib` uses for its CI.

This removes `PyMpiPoolExecutor` as a `Creator` attribute, changing the API, so not a patch; minor.

EDIT: After merging in the stack, this also removes the default `Creator.Executor`, so we now always specifically ask for ` ProcessPoolExecutor` or `executorlib.Executor` accordingly.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -12,4 +12,4 @@ jobs:
     uses: pyiron/actions/.github/workflows/tests-and-coverage.yml@actions-3.2.1
     secrets: inherit
     with:
-      extra-python-paths: tests tests/benchmark tests/integration tests/static tests/unit  # For pympipool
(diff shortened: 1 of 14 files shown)


liamhuber:
OSX tests started quickly, but timed out to failure

liamhuber:
Nope, still hanging. Unfortunately from the standard log it's not clear what test is doing it, and I can't reproduce it locally. @jan-janssen, thoughts?

liamhuber:
Hanging is just [CI related](https://github.com/pyiron/pyiron_workflow/issues/387#issuecomment-2248481380) and not a true failure

PR #391, opened 2024-07-24
Remove the extra python paths args from CI

liamhuber:
To see if executorlib has overcome this necessity that pympipool had


liamhuber:
No, `executorlib` still needs the paths

PR #410, opened 2024-08-08
[patch] `as_dataclass_node` pickle compatibility

### LINKED ISSUE #319: :bug: Dataclass nodes made from a decorator are not pickleable
The easy bit is that when coming from a decorator, they need the same flag as the macro and function nodes from a decorator `_class_returns_from_decorated_function` (although it's a class in this case, not a function, so I guess the flag needs a better name). Adding `cls._class_returns_from_decorated_function = dataclass` to `as_dataclass_node` is sufficient to resolve this, `__reduce__` then returns `factory._instantiate_from_decorated` and the problem is resolved in exactly the same manner as functions and macros.

The outstanding problem is that the underlying dataclass _also_ gets used by the output signal, and there it's an instance property rather than a class property, so pickle tries to deal with it in the `__getstate__` of the output channel and throws a `it's not the same object as` error.

I did a brute-force check and just setting `{dataclass node}.outputs.dataclass.type_hint = None` resolves this, so that really is the issue. In principle one could do this in the get- and set-state of dataclass nodes... but that's not a pattern I'm thrilled with.

Dataclass nodes instantiated separately from their dataclass instance pickle fine, of course. Maybe this is the better way to go anyhow, so that if/when you want to make new dataclasses by composition you don't wind up with something ugly like 

```python
@as_dataclass_node
@dataclass
class CompositeData(Parent.dataclass, Parent2.dataclass):
    pass
```

liamhuber:
Closes #319.

TODO:
- [x] Wait for the new version of `pyiron_snippets`
- [x] See if `as_node_function` and `as_macro_function` can use the same new syntax for `_reduce_imports_as`
- [x] Add `as_node_function` to the default creator now that it's working for pickle
- ~Wait for then bump `pyiron_base`~ I'll make the bump landing-page PR wait, but I want to collapse this stack.

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -12,7 +12,7 @@ dependencies:
 - pandas =2.2.2
 - pyiron_base =0.9.10
 - pyiron_contrib =0.1.17
-- pyiron_snippets =0.1.3
+- pyiron_snippets =0.1.4
 - python-graphviz =0.20.3
 - toposort =1.10
 - typeguard =4.3.0
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -12,7 +12,7 @@ dependencies:
 - pandas =2.2.2
 - pyiron_base =0.9.10
 - pyiron_contrib =0.1.17
-- pyiron_snippets =0.1.3
+- pyiron_snippets =0.1.4
 - python-graphviz =0.20.3
 - toposort =1.10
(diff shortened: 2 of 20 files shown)


liamhuber:
`pyiron_snippets` 0.1.4 is showing up on anaconda,  but must not be available here. This should be done modulo closing and re-opening to trigger a fresh download of the dependencies.

PR #422, opened 2024-08-12
[minor] Refactor storage

liamhuber:
Moving from three storage back-ends down to one, we have the opportunity to refactor the storage interface and apply lesson's learned on the first "alpha" pass at storage.


liamhuber:
End-of-day reminders to self:

- [ ] Format `checkpoint: Literal["pickle"] | StorageInterface | None` so that it can use something other than the default pickle back end
- [ ] Don't load by default for anything but `Workflow`. This is a subtlety I only noticed going through this refactoring -- if we auto-load everything you could have a macro that crashes at instantiation time because one of its children happens to have the same label as a saved node. As parent-most objects, it's safe to always greedily try loading new workflows
- [ ] Allow `any_storage_has_contents` to take a `StorageInterface` as an argument, and if it's not `None` try to delete there too. This is to accommodate loading at instantiation with a custom interface. Maybe the method can be private too.
- And more, this isn't exhaustive, just stuff I don't want to forget.

pyiron_workflow/storage.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+class TypeNotFoundError(ImportError):
+    """
+    Raised when you try to save a node, but importing its module and class give
+    something other than its type.
+    """
+
+
+class StorageInterface(ABC):
```
liamhuber, line 25: @pmrv this is the guy I was talking about

PR #427, opened 2024-08-13
[minor] Autoload

liamhuber:
Rename the `storage_backend` kwarg to `autoload`, and make a test showing how autoloading should _not_ be on by default (the exception is `Workflow` instances, which are parent-most anyhow, and thus safe and convenient to auto-load)


liamhuber:
All five Codacy "critical" failures are just it failing to grok that the decorator turns the functions into classes.

PR #435, opened 2024-08-22
[minor] Replace `HasToDict` mixin with `HasStateDisplay` mixin

liamhuber:
Functionality is similar -- the point is to get a nice JSON representation for child classes. But the public interface is different (hence minor flag), and in particular we don't use `to_dict` which might interfere with other pyiron activities. There's also some changes to exactly the display looks like, but since it's just a nice/useful thing for humans to look at, this is non-critical and can be improved with patches as needed.

### DIFF
--- a/pyiron_workflow/channels.py
+++ b/pyiron_workflow/channels.py
@@ -14,8 +14,8 @@ import inspect
 
 from pyiron_snippets.singleton import Singleton
 
-from pyiron_workflow.mixin.has_interface_mixins import HasChannel, HasLabel, UsesState
-from pyiron_workflow.mixin.has_to_dict import HasToDict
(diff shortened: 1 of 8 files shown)


liamhuber:
Coveralls failure: 
```
🚀 Posting coverage data to https://coveralls.io/api/v1/jobs
⚠️ Internal server error. Please contact Coveralls team.
```

I'll just try again in a minute

liamhuber:
> I'll just try again in a minute

Screen Shot 2024-08-22 at 13 41 56

Negative, [coveralls status](https://status.coveralls.io/) shows the API having problems for lots of people. This is a merge to a branch so I won't even worry about this.

liamhuber:
Codacy, on the other hand, actually caught an unused import. Good job, codacy.

PR #443, opened 2024-09-04
[patch] Obfuscate imports in the standard node library

liamhuber:
So they don't show up in tab completion menus -- _only_ the relevant node classes show up now!

@samwaseda, this idea came to me during our conversation on Monday. I don't think it's something that should be _necessary_ for people writing a node module, but maybe it's a best practice? It's a bit of a pain to write it this way, but it is the dead-simplest way to get the tab completion menu for the module to show only the nodes, which is a clear win. Thoughts?

### DIFF
--- a/pyiron_workflow/nodes/standard.py
+++ b/pyiron_workflow/nodes/standard.py
@@ -2,19 +2,26 @@
 Common-use nodes relying only on the standard library
 """
 
-from __future__ import annotations
+from __future__ import annotations as _annotations
 
-import random
-import os
-from pathlib import Path
-import shutil
-from time import sleep
+import random as _random
+import os as _os
+from pathlib import Path as _Path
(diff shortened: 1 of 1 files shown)


liamhuber:
An alternative would be to have an `pyiron_workflow.api` module and in `pyiron_workflow.__init__.py` just `from pyiron_workflow.api import *`. This could have submodules like `pyiron_workflow.api.standard_nodes`, where in the submodule we `from pyiron_workflow.nodes.standard import (...)` just grabbing the nodes we want as public.

I actually like the `api` approach a bit more as it's clearer/more rigorous. In both cases there's some typing legwork to do which is annoying. The big downside I see to the `api` approach is writing a new `pyiron_workflow.nodes.standard` class and then just simply forgetting to expose it in the API...

PR #444, opened 2024-09-05
[patch] List-like output for for-loops

### LINKED ISSUE #440: QoL: have `For` nodes return iterated output by list
And not _only_ by dataframe. We wind up very often jumping through hoops to unpack the results. I think the dataframe is exactly the response we want from the `.iter` and `.zip` methods, but when manually constructing a `for_node`, it seems like having separate output channels full of list-like data for each iterated output will be more useful. One can always recreate a dataframe from this with the `InputToDataframe` transformer node, but I also think we don't need to throw the dataframe representation away -- `for_node` is a function defining a new class, so we are free to modify output channels at that stage and can have a `output_dataframe: bool = False` option.

@JNmpi @ligerzero-ai any thoughts? Han, I'll ping you for review when I finally get a PR for this.

liamhuber:
WIP
- ~Refactor~ I'll come back if inspiration strikes, but otherwise I want to press on with features
- [x] Tests
- [x] Demo in the ~deepdive~ quickstart

Closes #440 

@ligerzero-ai, I still need to sand down the rough edges, but I hope this gets 70% of the way to resolving the issues you ran into when making the equation of state stuff. 

Before, because the for-loop was returning a dataframe, accessing the underlying dataclass attributes was a pain in the butt and was easiest to accomplish with custom logic

```python

@Workflow.wrap.as_dataclass_node
class SomeData:
    number: int = 42
    word: str = "towel"

wf = Workflow("data_access")

wf.loop = Workflow.create.for_node(
    body_node_class=SomeData,
    iter_on=("number",),
    number=[1, 2, 3],
    word="go",
)

@Workflow.wrap.as_function_node("just_one_field")
def Unpack(data):
    return [d.number for d in data["dataclass"].tolist()]


wf.unpacked = Unpack(wf.loop.outputs.df)

out = wf()
out
>>> {'unpacked__just_one_field': [1, 2, 3]}
```

Now, the same thing can be accomplished using built-in tools, in this case a second for-loop for the attribute access:

```python
@Workflow.wrap.as_dataclass_node
class SomeData:
    number: int = 42
    word: str = "towel"

wf = Workflow("data_access")

wf.loop = Workflow.create.for_node(
    body_node_class=SomeData,
    iter_on=("number",),
    number=[1, 2, 3],
    word="go",
    output_as_dataframe=False,
)

wf.unpacked = Workflow.create.for_node(
    body_node_class=Workflow.create.standard.GetAttr,
    iter_on=("obj",),
    obj=wf.loop.outputs.dataclass,
    name="number",
    output_as_dataframe=False,  # For backwards compatibility, dataframes are still default!
)

out = wf()
out
>>> {
...     'loop__number': [1, 2, 3],
...     'unpacked__obj': [
...         SomeData.dataclass(number=1, word='go'), 
...         SomeData.dataclass(number=2, word='go'),
...         SomeData.dataclass(number=3, word='go')],
...     ],
...     'unpacked__getattr': [1, 2, 3]
... }
```

At the moment we still need the extra step to unpack the dataclass. I can anticipate some feedback that this is still annoying, but at least for the moment I want to move one step at a time here. Basically it's not immediately clear to me whether we should be building up a bunch of infrastructure to help users unpack their dataclasses all over the place, or whether it's a better practice to leave them packages and encourage node packages that use dataclasses to use them throughout the node package (e.g. with multiple dispatch to allow dataclass _or_ explicit input). 

In the meantime I'm content that the extra step is at least "standard" stuff and not a custom node definition. I have some ideas for tightening the for-loop syntax though -- e.g. at least `iter_on` (and `zip_on`) should be able to take a single string instead of a tuple of length 1...


ligerzero-ai:
Thanks a lot Liam, I'll have a look at it over the weekend.

liamhuber:
From the updated quickstart:

```python
wf = Workflow("listlike_for_loop")
wf.first_loop = Workflow.create.for_node(
    body_node_class=Workflow.create.standard.Add,
    iter_on=("other",),
    obj=1,
    other=[1, 2, 4],
    output_as_dataframe=False,  # This is what's different
)
wf.second_loop = Workflow.create.for_node(
    body_node_class=Workflow.create.standard.Multiply,
    iter_on=("other",),
    obj=2,
    other=wf.first_loop.outputs.add,   # This is the nice part it enables!
    output_as_dataframe=False,
)
wf()
>>> {
...     'first_loop__other': [1, 2, 4],
...      'second_loop__other': [2, 3, 5],
...      'second_loop__mul': [4, 6, 10]
... }

```

Where we see the (in this case only) looped output `mul` as a channel of list data, and since they aren't connected to anything, the "what input got this output" `other` channels for both loops.

In the future I can see making the list-like output default and forcing people to ask for the dataframe.

PR #451, opened 2024-09-10
[patch] Let pint-hinted input values pass through type hinting

liamhuber:
By checking the actual datatype associated with their "magnitude".

I'm also boldly going to try setting the lower bound to the last version and hope it works.

Motivated by https://github.com/pyiron/elaston/pull/45


samwaseda (approved):
Yep this looks perfect to me :)

PR #458, opened 2024-09-18
Restarting process

liamhuber:
To work with long-duration nodes on executors that survive the shutdown of the parent workflow/node python process (e.g. `executorlib` using slurm), we need to be able to tell the run paradigm to serialize the results, and to try to load such a serialization if we come back and the node is running.

This introduces new attributes `Node.serialize_results` to trigger the result serialization, and a private `Node._do_clean` to let power users (i.e. me writing the unit tests) stop the serialized results from getting cleaned up automatically at read-time. 

Under the hood, `Node` now directly implements `Runnable.on_run` and `Runnable.run_args` leveraging the new detached path from #457 to make sure that each run has access to a semantically relevant path for writing the temporary output file (using cloudpickle). Child classes of `Node` implement new abstract methods `Node._on_run` and `Node._run_args` in place of the previous `Runnable` abstract methods they implemented.

TODO:
- [ ] ~Figure out how to get the node/parent workflow to checkpoint itself _after_ it has set its status to "running" when it is going to rely on serialization~ rebase this onto `main`; since breaking the `Runnable.run` cycle down, getting a right-before-running-checkpoint should be quite easy
- [ ] Try it out on cmmc to make sure it actually works in production
- [ ] Document it, probably in the deepdive

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -492,14 +492,13 @@ class Node(
         if self.parent is not None:
             self.parent.register_child_starting(self)
 
-        foo = super().run(
+        return super().run(
             check_readiness=check_readiness,
             force_local_execution=force_local_execution,
             _finished_callback=(
                 self._finish_run_and_emit_ran if emit_ran_signal else self._finish_run
             ),
         )
-        return foo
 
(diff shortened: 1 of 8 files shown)


liamhuber:
Superseded by #476

PR #459, opened 2024-09-19
Remove redundant channel definition

samwaseda:
It's defined above

### DIFF
--- a/pyiron_workflow/draw.py
+++ b/pyiron_workflow/draw.py
@@ -113,7 +113,6 @@ class _Channel(WorkflowGraphvizMap, ABC):
         self._parent = parent
         self._name = self.parent.name + local_name
         self._label = local_name + self._build_label_suffix()
-        self.channel: WorkflowChannel = channel
 
         self.graph.node(
             name=self.name,


liamhuber (changes requested):
Rather let's remove the other one -- then we have privates first then publics, which is not some rule but generally more consistent with how I've been writing the rest of the library, and we keep the one with the type hint

liamhuber:
> But isn't it more common to define the types only for the function args?

Usually -- crux of the issue is that I'm mis-remembering something. In the following:

```
class Foo:
    def __init__(self, x: int, y: int):
        self.x = x
        self.y: int = y


f = Foo(0, 1)
f.x = "breaks init hint"
f.y = "breaks class hint"
```

I thought my IDE would complain about the `f.y` assignment but let the `f.x` pass. Turns out it doesn't complain about either 🤦‍♂️ 

I can still see a space for hinting the attribute directly when the init hint differs from the reality of the attribute, e.g.:

```
class Bar:
    def __init__(self, x: list | None):
        self.x: list = [1, 2, 3] if x is None else x
```

liamhuber (approved):

samwaseda:
Yeah I also don't really understand what type hinting inside the function is good for. I guess we can think about it once we figure out what we can do with it.

liamhuber:
I swear I've seen my IDE respond usefully to it -- probably to an attribute that wasn't in the init signature to start with? But I couldn't reproduce anything in the three minutes I had to play with it

PR #460, opened 2024-09-19
change style depending on the state

samwaseda:
In the `draw` method I find it extremely difficult to figure out which (especially input) parameters still have to be defined, and I'm sure I'm not the only one feeling this way, but I'm not really sure what would be the best way to distinguish different states. In this example, only `Add__y` has no color, because it's not connected and no value is assigned:

```python
from pyiron_workflow import Workflow, function_node

def add(x, y):
    output = x + y
    return output

def mul(x, y=2):
    output = x * y
    return output

wf = Workflow("draw")
wf.Add = function_node(add, x=1)
wf.Mul = function_node(mul, x=wf.Add)

wf.draw()
```

Screenshot 2024-09-19 at 13 21 45

I find it pretty nice because you can immediately see that `Add__y` has to get a value or be connected. But this example is just a suggestion, which I don't consider to be the best possible product. So let me know what you think. @liamhuber @JNmpi @Tara-Lakshmipathy.

### DIFF
--- a/pyiron_workflow/draw.py
+++ b/pyiron_workflow/draw.py
@@ -174,7 +174,7 @@ class DataChannel(_Channel):
     @property
     def style(self) -> str:
         if len(self.channel.connections) == 0 and self.channel.value is NotData():
-            return "dashed"
+            return "bold"
         return "filled"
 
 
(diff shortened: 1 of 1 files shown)


liamhuber (approved):
Awesome! If possible I would like the line to be thicker or for there to be an 0.1 alpha fill or something since the line can be a bit hard to see -- but this is already enough of an improvement that this can be merged as-is if that's not fast to fix

PR #462, opened 2024-09-19
[patch] change color when failed

### LINKED ISSUE #92: Nodes in a failed state should be glaringly obvious in the graph draw
Maybe outlined in bright red or something.

samwaseda:
Closes #92

```python
from pyiron_workflow import Workflow, function_node

def add(x, y):
    output = x + y
    return output

def mul(x, y=2):
    raise ValueError
    output = x * y
    return output

wf = Workflow("draw")
wf.Add = function_node(add, x=1, y=2)
wf.Mul = function_node(mul, x=wf.Add)

wf.run()
```

Screenshot 2024-09-20 at 08 25 22

Btw. does it have a particular reason that node has a color?

### DIFF
--- a/pyiron_workflow/draw.py
+++ b/pyiron_workflow/draw.py
@@ -17,16 +17,29 @@ if TYPE_CHECKING:
 
 
 def directed_graph(
-    name, label, rankdir, color_start, color_end, gradient_angle, size=None
+    name,
+    label,
+    rankdir,
+    color_start,
+    color_end,
+    gradient_angle,
+    size=None,
+    style="filled",
+    fillcolor_start=None,
+    fillcolor_end=None,
 ):
     """A shortcut method for instantiating the type of graphviz graph we want"""
+    if fillcolor_start is None:
+        fillcolor_start = color_start
+    if fillcolor_end is None:
+        fillcolor_end = color_end
     digraph = graphviz.graphs.Digraph(name=name)
     digraph.attr(
         label=label,
         compound="true",
         rankdir=rankdir,
-        style="filled",
-        fillcolor=f"{color_start}:{color_end}",
+        style=style,
+        fillcolor=f"{fillcolor_start}:{fillcolor_end}",
         color=f"{color_start}:{color_end}",
         gradientangle=gradient_angle,
         fontname="helvetica",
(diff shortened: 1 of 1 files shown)


liamhuber (commented):
I really liked how for the present/missing data you did just the outline -- could we port that to doing the outline in red here and leaving the fill the usual colour?

liamhuber:
> Btw. does it have a particular reason that node has a color?

The end result should be that nodes of different types should have different colours (macros, function nodes, etc.) to improve visual clarity.

From a purity perspective, this should absolutely _not_ be a property of the node -- the `draw` method (or `pyiron-xyflow`) should handle it. But from a practical implementation perspective, that would mean that the drawer has its own colour map and is constantly doing `isinstance` checks to see which colour should be used -- grabbing `node.color` is much faster. I'm totally ok either route -- pure or practical, but practical is what there is right now.

samwaseda:
> From a purity perspective, this should absolutely _not_ be a property of the node -- the `draw` method (or `pyiron-xyflow`) should handle it. But from a practical implementation perspective, that would mean that the drawer has its own colour map and is constantly doing `isinstance` checks to see which colour should be used -- grabbing `node.color` is much faster. I'm totally ok either route -- pure or practical, but practical is what there is right now.

ok it also sounds to me like we should have a possibility to export levels, and it sounds to me like it's something that `pyiron_workflow` should be able to do. I actually kinda have the feeling that the drawing module should also be better separated. I'll look into it at some point.

samwaseda:
> I really liked how for the present/missing data you did just the outline -- could we port that to doing the outline in red here and leaving the fill the usual colour?

I updated it and also added an example in the top comment.

liamhuber (approved):

liamhuber:
> we should have a possibility to export levels

Can you clarify this? IMO it's not as clear as "levels" of node -- that makes sense for function vs macro, but not so much for function vs transformer, or for loop vs macro.

liamhuber:
> I actually kinda have the feeling that the drawing module should also be better separated. 

I see what you're saying, but I will disagree; I think it's nice to have some sort of built-in way of visualizing the nodes that comes with the library. It's a common enough operation that I'm disinclined to force people to separately install a second tool to get it done.

PR #468, opened 2024-09-22
Get close matches

samwaseda:
```python
wf = Workflow("tesile")
wf.Composition = standard_nodes.UserInput("Aluminum")
wf.BoxSize = standard_nodes.UserInput(3 * [1.0e-5])
wf.Seeds = function_node(...)
wf.Grid = function_node(...)
wf.Elasticity = function_node(...)
wf.Plasticity = function_node(...)
wf.Phase = function_node(...)
wf.Homogenization = function_node(...)
wf.Material = function_node(generate_material, elements=wf.Compositions)
```

The previous error message in this situation was: `Could not find attribute Compositions on tesile (Workflow) or among its children (dict_keys(['Composition', 'BoxSize', 'Seeds', 'Grid', 'Elasticity', 'Plasticity', 'Phase', 'Homogenization'])).`. It was sort of difficult to figure out which one I wanted to say for this large number of possibilities. So I added `get_close_matches` to append the message: `Did you mean 'Composition' and not 'Compositions'?`

Also, in general I find the error messages in `pyiron_workflow` difficult to understand. It must be difficult for @liamhuber to figure it out alone so I'm gonna make suggestions in the upcoming weeks.

### DIFF
--- a/pyiron_workflow/io.py
+++ b/pyiron_workflow/io.py
@@ -77,7 +77,7 @@ class IO(HasStateDisplay, ABC):
             # Raise an attribute error from getattr to make sure hasattr works well!
             raise AttributeError(
                 f"Could not find attribute {item} on {self.__class__.__name__} object "
-                f"nor in its channels ({self.labels})"
+                f"nor in its channels ({self.labels})."
             )
 
     def __setattr__(self, key, value):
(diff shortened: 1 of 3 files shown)


samwaseda:
The same thing could be used also elsewhere (like io?)

liamhuber (approved):
Beautiful! As a += this is a strict improvement

liamhuber:
> The same thing could be used also elsewhere (like io?)

Yes, I imagine there's a few places where it's useful, but I couldn't name them all off the top of my head. When it's used as here to strictly add to the information returned, I am totally behind this.

PR #469, opened 2024-09-22
Update error message of IO when an inexistent arg is used

samwaseda:
Code:
```python
def f(a, b, c):
    return a

function_node(f, a=1, b=2, d=3)
```

Previous error message:
`ValueError: Tried to set input ['a', 'b', 'd'], but one or more label was not found among available inputs: ['a', 'b', 'c']`

New error message:
`ValueError: {'d'} not found among available inputs: ['a', 'b', 'c']`

### DIFF
--- a/pyiron_workflow/io.py
+++ b/pyiron_workflow/io.py
@@ -420,7 +420,8 @@ class HasIO(HasStateDisplay, HasLabel, HasRun, ABC):
         for k, v in kwargs.items():
             self.inputs[k] = v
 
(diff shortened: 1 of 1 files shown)


liamhuber (approved):
Super, another pure improvement 🚀 

Looks like the new method can be `staticmethod`; please verify and make it one if that's the case

samwaseda:
> Looks like the new method can be `staticmethod`; please verify and make it one if that's the case

Yep you are right. I changed it

Issue #456, opened 2024-09-18
Syntax ideas and suggestions for for_loop and executors in pyiron_workflows


JNmpi:
@liamhuber, below are a couple of ideas and thoughts to make the syntax for controls such as for loops and executors in pyiron_workflows easier. They are not intended as a final solution but more to start a discussion of how features that are considered by users to be (too) complex can be made easier and more intuitive.

#### Extend .iter to macro programming

    wf.nodes = wf.create.my_path.MyNode(a=1, b=2).iter(element=['Al', 'Fe'])

This should be easy to achieve by taking all relevant info to perform the present definition, i.e.,

    self.nodes = Workflow.create.for_node(
        body_node_class=Workflow.create.my_path.MyNode, # extract class name from node
        iter_on="element",
        obj='Al',
        other=['Fe']
    )

This would be much more intuitive. During the workshop people were actually trying such a construct.    

#### Direct approach

    wf.nodes = wf.create.my_path.MyNode(a=1, b=2).server(cores=4).iter(element=['Al', 'Fe', 'Pb']).server(workers=3, queue='test')

Run 3 MyNode jobs simultaneously on 4 cores each, i.e., the queue test is assumed/told to provide 12 cores. This notation can be easily extended to include additional control variables (e.g. for controling the log level, gpu vs cpu), etc.    

#### Indirect approach via workflow

    wf = Workflow('my_workflow').server(cores=12, queue='test')

    # run interactively (no submission to queue)
    wf.interactive_node = wf.create.my_path.MyNode(a=1, b=2) 

    # run by submitting to queue (note: the number of workers is not needed, is obtained from 12 cores/ 4 cores per node)
    wf.expensive_nodes = wf.create.my_path.MyNode(a=1, b=2).server(cores=4, to_queue=True).iter(element=['Al', 'Fe', 'Pb'])

    wf.run()

Alternative formulation (both could be supported)
    
    wf = Workflow('my_workflow')

    # run interactively (no submission to queue)
    wf.interactive_node = wf.create.my_path.MyNode(a=1, b=2) 

    # run by submitting to queue (note: the number of workers is not needed, it is obtained from 12 cores/ 4 cores per node)
    wf.expensive_nodes = wf.create.my_path.MyNode(a=1, b=2).server(cores=4, to_queue=True).iter(element=['Al', 'Fe', 'Pb'])

    wf.server(cores=12, queue='test').run()

### Further extensions of the server 

    server(
[...]

PR #472, opened 2024-09-24
[minor] Remove `StaticIO.iter` and `.zip` shortcuts

liamhuber:
and update quickstart accordingly.

@JNmpi, from #456 it sounds like not only would you like to have different behaviour for `node.iter(...)`, but also that the current behaviour was actively confusing at the hackathon. I would suggest then that we just remove the current behaviour; this avoids confusion immediately, and in the long run frees up `.iter` and `.zip` in the node namespace for behaviour that better represents what you want. By removing it now, re-introducing the improved behaviour will avoid breaking backwards compatibility.


JNmpi:
This proposed change goes in the opposite direction of what I wanted to achieve. The main issue is that the current for_loop syntax was considered by many to be too complex and non-intuitive, and there was concern that such constructs would scare people away. The .iter statement, although only briefly discussed, was seen as an easy to use approach. In fact, when the benefits of the workflow approach were highlighted, this was one of the killer features that immediately won people over. So I would really like to keep it, or extend it based on the discussion we had in [#456](https://github.com/pyiron/pyiron_workflow/issues/456). Rather than removing such a feature, I would like to push the syntax in a way that feels comfortable and intuitive to users. I am fully aware that this is not an easy path, and will require several iterations to arrive at constructs that are easy to use without introducing unwanted or unexpected side effects. But to be successful, this is the path we must take.

PR #479, opened 2024-09-27
[patch] Add a read-only example of running nodes on HPC

liamhuber:


jan-janssen:
I really like this step towards an HPC interface, I am just a bit surprised you are not using https://github.com/pyiron-dev/remote-executor or https://github.com/pyiron/pysqa . Is there any specific reason for re-implementing the queuing system interface?

jan-janssen:
I updated the `pysqa` interface to support dynamic definition of the submission script in `pysqa=0.2.0`, I hope this simplifies the integration in a functional approach, like the one here in `pyiron_workflow`: https://pysqa.readthedocs.io/en/latest/example_queue_type.html

liamhuber:
Yes, at the surface level it looks like that would directly replace all the sbatch script writing in the example here! I won't have a chance to test it immediately though.

PR #481, opened 2024-09-29
[patch] Catch keyboard interrupts in `Runnable._run`

liamhuber:
I didn't realize they weren't caught by `Exception`; this way the keyboard interrupt still hits the `Runnable._run_finally`, and in particular `Node` still generates a recovery file

### DIFF
--- a/pyiron_workflow/mixin/run.py
+++ b/pyiron_workflow/mixin/run.py
@@ -207,7 +207,7 @@ class Runnable(UsesState, HasLabel, HasRun, ABC):
         if executor is None:
             try:
                 run_output = self.on_run(*on_run_args, **on_run_kwargs)
-            except Exception as e:
+            except (Exception, KeyboardInterrupt) as e:
                 self._run_exception(**run_exception_kwargs)
                 self._run_finally(**run_finally_kwargs)
                 if raise_run_exceptions:


liamhuber:
For instance, we can slightly modify the "recovering from failure" example in the HPC notebooks so there's a slow-running node on an executor that won't fail. 

```python
import pyiron_workflow as pwf

wf = pwf.Workflow("test")
wf.a = pwf.standard_nodes.UserInput(1)
wf.b = pwf.standard_nodes.UserInput("two")
wf.slow = pwf.standard_nodes.Sleep(5)
wf.c_fails = wf.a + wf.b  # Type error
wf.d_if_success = pwf.standard_nodes.UserInput(0)
wf.d_if_failure = pwf.standard_nodes.UserInput("But what's the question?")
wf.e_fails = pwf.standard_nodes.Add(wf.d_if_failure, 42)  # Type error

wf.a >> wf.b >> wf.c_fails >> wf.d_if_success
wf.b >> wf.slow
wf.c_fails.signals.output.failed >> wf.d_if_failure >> wf.e_fails
wf.starting_nodes = [wf.a]
wf.automate_execution = False

with pwf.Workflow.create.ProcessPoolExecutor() as exe:
    wf.slow.executor = exe
    wf()
```

If we then run this directly in a cell and cancel it, we now get the desired behaviour of generating recovery file. This can be reloaded and drawn:

recovered

PR #483, opened 2024-10-02
Target actions branch

liamhuber:
To test the new miniforge variant default

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   pyiron:
-    uses: pyiron/actions/.github/workflows/push-pull.yml@actions-3.3.1
+    uses: pyiron/actions/.github/workflows/push-pull.yml@forge-variant
     secrets: inherit
     with:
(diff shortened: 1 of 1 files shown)


liamhuber:
```
Ensuring installer...
  Can we use bundled Miniconda?
  Can we download a custom installer by URL?
  Can we download Miniforge?
  ... will download Miniforge.
  Will fetch Miniforge latest from https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge-Linux-x86_64.sh
  Ensuring Installer...
  Checking for cached Miniforge@latest...
  Did not find Miniforge-Linux-x86_64.sh latest in cache
Error: Unexpected HTTP response: 404
```

For Pete's sake, I'm not allowed to use the `Mambaforge` variant anymore, but `latest` is throwing errors, even though the `https://github.com/conda-forge/miniforge/releases/download/latest` is a valid page and it, in turn, contains `https://github.com/conda-forge/miniforge/releases/download/24.7.1-2/Mambaforge-Linux-x86_64.sh`

PR #486, opened 2024-10-08
add colors for gui

Tara-Lakshmipathy:


pyiron_workflow/nodes/composite.py
```diff
@@ -423,6 +423,11 @@ def color(self) -> str:
         """For drawing the graph"""
         return SeabornColors.brown
 
+    @property
+    def gui_color(self) -> str:
+        """For nodes in the gui"""
+        return "#eacf9f"
```
samwaseda, line 429: I find it nice that @liamhuber so far mostly followed the convention of naming the color and return the variable. Do you maybe want to do that as well?

samwaseda (commented):
Can you add an example figure?

pyiron_workflow/nodes/composite.py
Tara-Lakshmipathy: I added variables for the colors, but did not define them in the Seaborn class (which is in pyiron_snippets). I think the Seaborn colors follow some pre-defined color palette . Is that OK?

Tara-Lakshmipathy: And where would you like the example picture?

samwaseda: In the PR comment. Actually you might need to explain either in the code (better) or in the PR how this property is used.

Tara-Lakshmipathy:
Gui6

The properties are used in `pyiron_xyflow` [here](https://github.com/pyiron/pyiron-xyflow/blob/main/python/reactflow.py#L148) to color the nodes.

samwaseda:
can you maybe also say real quick what was the problem without having `gui_color`?

pyiron_workflow/nodes/composite.py
```diff
@@ -423,6 +423,12 @@ def color(self) -> str:
         """For drawing the graph"""
         return SeabornColors.brown
 
+    @property
+    def gui_color(self) -> str:
+        """"""For nodes in the gui, pyiron-xyflow in reactflow.py""""""
```
samwaseda, line 428:
```suggestion
        """For nodes in the gui, pyiron-xyflow in reactflow.py"""
```

pyiron_workflow/nodes/transform.py
```diff
@@ -29,6 +29,12 @@ def color(self) -> str:
         """For drawing the graph"""
         return SeabornColors.blue
 
+    @property
+    def gui_color(self) -> str:
+        """"""For nodes in the gui, pyiron-xyflow in reactflow.py""""""
```
samwaseda, line 34:
```suggestion
        """For nodes in the gui, pyiron-xyflow in reactflow.py"""
```

Tara-Lakshmipathy: Ooops :P

Tara-Lakshmipathy:
> Ok that's a valid concern. But I have a different concern now: Since we rely on auto-complete as well as `__setattr__` and `__getattr__`, it might be preferable to keep the number of attributes as low as possible. Maybe the easiest solution for now is to change `color` by `get_color` and take an argument for GUI. However, I have the feeling that the real source of the problem is the fact that the node objects (nodes, macros etc.) have colors. Maybe it's really time to think about changing this.
> 
> Let's wait for @liamhuber and see what he says

There may be more such attributes for the automated positioning of nodes in the gui in the future. If not in the specific types of nodes (e.g., function.py``) then at least in `node.py`. So, maybe it's already worth discussing where such attributes should go.

liamhuber:
I am definitely against introducing a new field for the gui color -- the core infrastructure shout have no idea the gui exists, and @samwaseda is correct about tab completion menu crowding.

I have no objection to updating the existing color attributes and that should solve the problem.

I am ok totally getting rid of the color node attribute, but this information can't be destroyed -- only moved. _Somewhere_ we should have a table relating classes to colors, and a function for getting most-specific class color from this table given the class. Pros: nodes don't need to know about colors and then they wouldn't, and draw and the guy could use the same infrastructure with different color tables. Cons: more of a pain to maintain as such color tables may need to be updated (maybe in multiple places) if a new node is introduced. Overall I still think this is the best attack.

samwaseda:
> the core infrastructure shout have no idea the gui exists

And for this very reason I find it actually surprising that there are colours attached to the nodes in the first place. For me a helpful distinction would be whether there are more nodes underneath, which in the case of `pyiron_workflow` coincides with `node` vs. `macro`, but I guess we can also do the same distinction by whether it has children or not?

One way or other, I think there should be a cleaner interface between `node` and `draw`. Ideally `pyiron_workflow` should be able to export an entire graph first in a simpler format like `dict`, which is taken over by `draw.py`. And from my point of view, it should be entirely up to `draw.py` to figure out how to plot it.

For this PR, if it's all about darkness, we can maybe define something similar to [`lighten_hex_color`](https://github.com/pyiron/pyiron_workflow/blob/d8728f13d5a609970ca1da6abff714bef531db87/pyiron_workflow/draw.py#L83) for now on the GUI side?

Tara-Lakshmipathy:
Is there an attribute which gives me the type of the node - function, composite etc.? I feel this would be useful I'm a variety of situations. On the S-Bahn for the next 50 minutes, but can't wait that long to check for myself 😆.

liamhuber:
> But I think having an attribute that returns the type of node directly...

I disagree -- there is a straightforward pythonic way of finding what type of `node` we have and it is `type(node)` or `isinstance(node, ThingIAmCheckingAgainst)`. This is already super straightforward and I don't think we should reinvent the wheel/create more 

> ...would be a neat convenience feature for a bunch of front-end applications, especially when an external tool wants to interface with 'pyiron_workflow' without knowing about the internal python classes used in the workflow code.

Here I agree, but the way I suggest to do it is as I mentioned earlier: `pyiron_workflow.draw`, which is embedded right in the module and knows all the core classes, should know how to map those core classes to nice default colours. This is convenient without stepping on python's toes, and should cover 90%+ of use cases. If someone wants a non-standard colour for some particular node, nothing stops them from doing that, and whatever `pyiron_workflow.node.get_color(node: Node)` looks like, it should be flexible enough that they can extend it. E.g. perhaps it's like `pyiron_workflow.draw.get_color(node: Node, color_map: Optional[dict]: None)` and we can also provide `pyiron_workflow.draw.color_map: dict[type[Node], str]` that they could import, extend/override, then pass to `get_color`.

So I'm for convenience, but against alternate routes for standard language stuff.

Tara-Lakshmipathy:
If we have `get_color` as opposed to the `color` we have now, then I would be in agreement. And since this pull request is about colors, I don't want to deviate from that topic too much.  But I agree with @samwaseda mentioned earlier. I was surprised to find color in these files. Initially I expected to find it in `draw.py` as well. So if we all agree to attack this by getting rid colors from the nodes themselves, then count me in.

PR #499, opened 2024-11-21
Define error e if not defined

samwaseda:
See lines changed


liamhuber (approved):
Is this just to shut your IDE up, or did you actually manage to trigger an edge case where this is relevant? Making the IDE happy is a sufficient reason for me, but since we generically catch `Exception`, the only way we could ever get to the old 135 _without_ `e` is if `for suffix, save_method in attacks:; p = filename.with_suffix(suffix)` raised an error -- I don't think this is possible with the current code, i.e. we'd need to go back and change the definition of `attacks` or similar to be able to trigger an error here.

liamhuber:
No no, this is formally correct so I'm ok with merging it

PR #520, opened 2024-12-11
Draft: Ruff

XzzX:
Apply `ruff check` to the repository.


liamhuber:
Awesome, thanks! I skimmed through the unsafe changes on mobile. The vast majority look good, but there's a few where variables are switched to private versions, or assignments are removed completely, that I need to be on my desktop to go dig into the context to see if they're ok. Some of the unit tests seem to be failing in a meaningful way, so hopefully these are connected issues and resolving my red flag changes will trivially resolve the (non-ruff) test issues. I'm afraid I won't be able to dig into it until tomorrow afternoon.

XzzX:
> Awesome, thanks! I skimmed through the unsafe changes on mobile. The vast majority look good, but there's a few where variables are switched to private versions, or assignments are removed completely, that I need to be on my desktop to go dig into the context to see if they're ok. Some of the unit tests seem to be failing in a meaningful way, so hopefully these are connected issues and resolving my red flag changes will trivially resolve the (non-ruff) test issues. I'm afraid I won't be able to dig into it until tomorrow afternoon.

The unit test failures all happened because ruff removed unnecessarily complicated code. However, this was used to trigger things in pyiron_workflow which does not happen anymore. I reverted all the changes that broke tests and annoted them so they do not get changed in the future. Test CI is now green.

XzzX:
Now come the more complicated issues.

How do you feel about [F401](https://docs.astral.sh/ruff/rules/unused-import/) and the top-level pyiron_workflow import?

Error: pyiron_workflow/create.py:47:5: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks also seems like an issue we should look into.

I will do further fixes by hand.

XzzX:
@liamhuber I think the rest of the errors is up to you to decide/fix.

liamhuber:
> How do you feel about [F401](https://docs.astral.sh/ruff/rules/unused-import/) and the top-level pyiron_workflow import?

The problem is they are only "unused" in a very technical sense that they are not further referenced in that file. In reality, they are used to specify what makes up the "public API" for the sake of semantic versioning. AFAIK there is no standard for how to do this, but however it is done it should be clear. I find this is both technically clear (there is a comment in the file stating that the imports here make up the public API) and practically clear (this is what's easiest for the public to access). So in the absence of a better way to accomplish this goal, these imports should not be removed.

liamhuber:
> Error: pyiron_workflow/create.py:47:5: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks also seems like an issue we should look into.

This is a good point. I would need to dig into it deeper to say something with absolutely certainty, but after a ten minute investigation I'm at least cautiously optimistic: it sounds like the concern is that the cache prevents the `self` argument of the cached methods from being released by the garbage collector. In our particular case, these ruff errors are all coming from `pyiron_workflow.create.Creator`, which is a singleton that gets referenced by every `Composite` instance. So we're only preserving a reference to a single instance of this thing, and it's not likely we ever want to garbage collect it until we're totally done working with `pyiron_workflow` anyhow.

XzzX:
> > How do you feel about [F401](https://docs.astral.sh/ruff/rules/unused-import/) and the top-level pyiron_workflow import?
> 
> The problem is they are only "unused" in a very technical sense that they are not further referenced in that file. In reality, they are used to specify what makes up the "public API" for the sake of semantic versioning. AFAIK there is no standard for how to do this, but however it is done it should be clear. I find this is both technically clear (there is a comment in the file stating that the imports here make up the public API) and practically clear (this is what's easiest for the public to access). So in the absence of a better way to accomplish this goal, these imports should not be removed.

So which solution do you prefer? I think using the file for the public API is totally valid. I vote for the `__all__` solution. Second best option without a preference are silencing the warning or doing a redundant import alias.

XzzX:
> > Error: pyiron_workflow/create.py:47:5: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks also seems like an issue we should look into.
> 
> This is a good point. I would need to dig into it deeper to say something with absolutely certainty, but after a ten minute investigation I'm at least cautiously optimistic: it sounds like the concern is that the cache prevents the `self` argument of the cached methods from being released by the garbage collector. In our particular case, these ruff errors are all coming from `pyiron_workflow.create.Creator`, which is a singleton that gets referenced by every `Composite` instance. So we're only preserving a reference to a single instance of this thing, and it's not likely we ever want to garbage collect it until we're totally done working with `pyiron_workflow` anyhow.

Yes, the concerns are twofold. a) garbage collection might not work due to the captured self, like you pointed out and b) a separate cache will be built for every instance. Meaning, even though the function call might be exactly the same, if it is called from a different instance, it is not.

I personally do not like singletons at all and I do not like the idea of "we do not use it wrongly" as that might change without notice in the future. Therefore, I vote for fixing it.

XzzX:
> > Error: pyiron_workflow/create.py:47:5: B019 Use of `functools.lru_cache` or `functools.cache` on methods can lead to memory leaks also seems like an issue we should look into.
> 
> This is a good point. I would need to dig into it deeper to say something with absolutely certainty, but after a ten minute investigation I'm at least cautiously optimistic: it sounds like the concern is that the cache prevents the `self` argument of the cached methods from being released by the garbage collector. In our particular case, these ruff errors are all coming from `pyiron_workflow.create.Creator`, which is a singleton that gets referenced by every `Composite` instance. So we're only preserving a reference to a single instance of this thing, and it's not likely we ever want to garbage collect it until we're totally done working with `pyiron_workflow` anyhow.

As far as I can see you are using it to cache module imports? Is it really worth doing this?  
https://docs.python.org/3/reference/import.html#the-module-cache

liamhuber:
> As far as I can see you are using it to cache module imports? Is it really worth doing this?

There are also cases where we're instantiating a dictionary with modules as elements. I'm sure that's slower than the module cache access. I'm happy to abandon caching and use straight properties, but I would still do a quick timeit check on, e.g., `Workflow.create.transformer` with and without the caching to double-check that it is as cheap as we imagine.

liamhuber:
> So which solution do you prefer? I think using the file for the public API is totally valid. I vote for the `__all__` solution. Second best option without a preference are silencing the warning or doing a redundant import alias.

I would prefer silencing the error as it's the most explicit resolution to the problem -- i.e. the linter simply (and fairly) isn't sophisticated enough to understand our intent with the imports, so just let it know. I'm also open to redundant aliases. I have a strong preference against `__all__`, because it has the side effect of allowing people to `from pyiron_workflow import *` -- I don't want to read code that uses this feature, and since I get pinged to debug `pyiron_workflow` code I choose to simply disallow it 😝

liamhuber:
> if we de-singleton the creator then we _MUST NOT_ allow such a return [to a registration system], as it will make the shortcuts hell again.

As long as you're on board with this take, I would stack some PRs on this to:
- De-singleton classes in `create`
- Check the timing and (probably) remove the caching in `create`
- Touch up the remaining complaints and implement the formatting suggestions

XzzX:
> > So which solution do you prefer? I think using the file for the public API is totally valid. I vote for the `__all__` solution. Second best option without a preference are silencing the warning or doing a redundant import alias.
> 
> I would prefer silencing the error as it's the most explicit resolution to the problem -- i.e. the linter simply (and fairly) isn't sophisticated enough to understand our intent with the imports, so just let it know. I'm also open to redundant aliases. I have a strong preference against `__all__`, because it has the side effect of allowing people to `from pyiron_workflow import *` -- I don't want to read code that uses this feature, and since I get pinged to debug `pyiron_workflow` code I choose to simply disallow it 😝

I did it.

XzzX:
> > if we de-singleton the creator then we _MUST NOT_ allow such a return [to a registration system], as it will make the shortcuts hell again.
> 
> As long as you're on board with this take, I will stack some PRs on this to:
> 
> * [ ]  De-singleton classes in `create`
> * [ ]  Check the timing and (probably) remove the caching in `create`
> * [ ]  Touch up the remaining complaints and implement the formatting suggestions

Let's keep the singleton if this is what you want. I would prefer removing `Creator` and `Wrappers` and enforce "import-what-you-want`, i.e. the pattern you mentioned above.

Check timing, go for it.

Formatting, we do not need to include `ruff format`. As long as we do have a formatter in place, I am fine. And black is a good choice. At some point I would extend the black coverage to also tests. Unfortunately, black also does the ugly multiline string formatting. I agree with you the other indentation style looks nicer.

liamhuber:
> What is wrong with
>
> ```python
> from pyiron_workflow import Workflow
> import some_node_package as snp
> ...
> wf.some_node = snp.SomeNode(...)
> ```
> 
> To me, that is very clear and explicit.
> ```

💯. Nothing is wrong with it. This is how we do it and how I want to keep doing it!

liamhuber:
> I did it.

Awesome, thanks! I had imagined we needed to do a line-by-line ignore; that patch is nice and simple.

liamhuber:
> Let's keep the singleton if this is what you want. I would prefer removing `Creator` and `Wrappers` and enforce "import-what-you-want`, i.e. the pattern you mentioned above.

This is my preference and how I use it, because I already know what I'm looking for. One of the "pyiron" ideas is to lean heavily tab completion, and one way to do that is being able to import a single object that acts as a starting point object for all the built-in functionality (i.e. `Workflow`). It's not my preferred usage strategy, but I see why others like it and we're not likely to convince the whole team, so I think the lesser evil is to continue to support both approaches.

For "third party" tools, like individual node packages, I think we'll be able to keep everyone on the same page with "import what you want".

> Check timing, go for it.
 
👍 

> Formatting, we do not need to include `ruff format`. As long as we do have a formatter in place, I am fine. And black is a good choice. At some point I would extend the black coverage to also tests. Unfortunately, black also does the ugly multiline string formatting. I agree with you the other indentation style looks nicer.

Ok, I'm also content to stick with black. It doesn't complain about the multiline indentation as it is, which is my biggest complaint with ruff. That said, ruff highlighted a few changes I do prefer, so I will probably just run it locally every now and then and cherry pick the suggested changes that improve readability. For the CI we can stick with black.

XzzX:
> > Let's keep the singleton if this is what you want. I would prefer removing `Creator` and `Wrappers` and enforce "import-what-you-want`, i.e. the pattern you mentioned above.
> 
> This is my preference and how I use it, because I already know what I'm looking for. One of the "pyiron" ideas is to lean heavily tab completion, and one way to do that is being able to import a single object that acts as a starting point object for all the built-in functionality (i.e. `Workflow`). It's not my preferred usage strategy, but I see why others like it and we're not likely to convince the whole team, so I think the lesser evil is to continue to support both approaches.
> 
> For "third party" tools, like individual node packages, I think we'll be able to keep everyone on the same page with "import what you want".

But you do get almost the same tab completion with imports?!?
```python
import some_node_package as snp
wf.some_node = snp.<tab>
```
vs
```python
wf.register("some_node_package", "snp")
wf.some_node = wf.create.snp.<tab>
```
I do not really see the point.

> > Formatting, we do not need to include `ruff format`. As long as we do have a formatter in place, I am fine. And black is a good choice. At some point I would extend the black coverage to also tests. Unfortunately, black also does the ugly multiline string formatting. I agree with you the other indentation style looks nicer.
> 
> Ok, I'm also content to stick with black. It doesn't complain about the multiline indentation as it is, which is my biggest complaint with ruff. That said, ruff highlighted a few changes I do prefer, so I will probably just run it locally every now and then and cherry pick the suggested changes that improve readability. For the CI we can stick with black.

That is a misperception. `black` does not complain because you do not run it on the test directory ;) For `pyiron_workflow` `ruff format` only adds trailing commas to lists which I can see why you might want to have that.

liamhuber:
> But you do get almost the same tab completion with imports?!?
> 
> ```python
> import some_node_package as snp
> wf.some_node = snp.<tab>
> ```
>
> vs
> 
> ```python
> wf.register("some_node_package", "snp")
> wf.some_node = wf.create.snp.<tab>
> ```
> 
> I do not really see the point.

Yes, you're preaching to the choir. Leveraging the top-level `__init__` file to define the API gives exactly this advantage. This (or `from pyiron_workflow...`) is how I use the package myself, and I try to be consistent in using it that way for demos, examples, and use cases I touch (e.g. in `pyiron_nodes`) in an attempt to shift the user-culture towards what I see as a better approach.

But this is what I'm trying to communicate, the presence of an importable starting point _class_ that is equiped with (nearly) the same tab-completion capabilities is a _cultural feature expectation_ of our community of users. In my estimation, the culture has not yet shifted enough to simply axe this feature  without complaint. Maybe I'm wrong. I don't feel strongly enough to push for a step-change on this; if you do you're welcome to suggest it at the next meeting in January, and I'm happy to say I agree with you, but I'm not willing to fight anyone for the change.


> `black` does not complain because you do not run it on the test directory ;)

🤦‍♂️ 😂 derp, yep!



> For `pyiron_workflow` `ruff format` only adds trailing commas to lists which I can see why you might want to have that.

Yes, I like those. I also noticed a couple newlines and some other small stuff I liked. 

Black is sufficient, but actually it would probably be nice to extend the black to cover the tests too -- https://github.com/pyiron/actions/issues/137

liamhuber:
Once #526 and #527 are merged down, I'm happy and I think we've done everything discussed here.

PR #524, opened 2024-12-18
Draft: add black formatting for tests directory

XzzX:
Waiting for   
https://github.com/pyiron/actions/pull/138  
and  
https://github.com/pyiron/pyiron_workflow/pull/520


liamhuber:
I'm sick and my kid's school hasn't restarted yet, so I'm not going to get to resolving the conflicts yet. Easier will probably be to just cherry pick the top commit and then re-black.

Per your comment elsewhere (sorry, lost the ref), if black is going to complain about the multi-line strings, it might be time to just give up on that, despite the fact we both find our version somewhat more readable. The silver lining would be that we could abandon black entirely and just use the ruff formatting...

liamhuber:
Ah, sorry @XzzX, I tried to roll back head and merge main so that your commit touching `pyproject.toml` stayed intact, but evidently I failed. I don't feel it's a big enough deal to go back and try all over again, but just FYI I was trying to keep you credited for the change in the git history.

> [We need to find a way how to deal with multiline strings also in black.](https://github.com/pyiron/pyiron_workflow/pull/526#pullrequestreview-2527138214)

I'm ready to concede to the formatters on this this one. If `ruff` and `black` are both pushing the same thing, I don't have enough willpower to fight it. Are you ok with their slightly-less-readable strings?

I was briefly excited because they were both ok with the following:

```python
        self.assertLess(
            dt,
            1.1 * t,
            msg="""
                Expected the sleeps to run in parallel with minimal overhead (since 
                it's just a thread pool executor) -- the advantage is that the "
                constructors should survive (de)serialization
            """,
        )
```

(Also as `f"""`-strings.) But, while this is perfectly readable from the code perspective, it actually prints the error message with all the extra indentation 🤦‍♂️

PR #526, opened 2024-12-20
Slightly less ruff

liamhuber:
I got stuck in authentification hell, so I only had time to just get started; more to come.

Timing test:

```python

%%timeit
Workflow.create.transformer

```

With lru cache: `45.3 ns ± 0.584 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)`

Without: `46.6 ns ± 0.398 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)`

A completely inconsequential difference, so I got rid of the creator/wrapper caches. I'll check the remaining cache next.


liamhuber:
@XzzX, with these amendments I'm happy with #520. Basically I accepted most of the formatting, but ultimately I turned it off and we'll lean on black for now. Otherwise the only anti-ruff thing I did was keep the top-level `Workflow` import un-sorted, and tell it to _always_ ignore unused imports in `__init__.py` files (since we use these to specify APIs).

pyproject.toml
```diff
@@ -84,3 +84,6 @@ select = [
     "I",
 ]
 ignore = ["E501"] #ignore line-length violations
+
+[tool.ruff.lint.per-file-ignores]
+"__init__.py" = ["F401"]  # Ignore unused imports in init files -- we specify APIs this way
```
XzzX, line 89: I am unsure if the generalization here is necessary, but I am ok with it.

XzzX (approved):
We need to find a way how to deal with multiline strings also in black.

PR #532, opened 2025-01-06
Target the ruff branch and activate one of the two jobs

liamhuber:
Not for merging, just testing the CI.


liamhuber:
I expected the check to run, but not the sort-imports since it defaults to false -- yet there they both are...

liamhuber:
Ok, skipping black worked fine, and now both ruff jobs get skipped -- I think maybe I was looking at the ruff job and not the ruff subjob of the push-pull job when I thought they were running 🤦‍♂️ 

I had a typo in the flag upstream for `do-ruff-check`, so I'm not surprised it still gets skipped here -- I usually (always?) need to close and re-open PRs to convince the PR to grab the freshest upstream CI stuff.

liamhuber:
I had pushed only one of the two pyiron/actions ruff branches -- all is now as-expected here, and I am confident https://github.com/pyiron/actions/pull/139 will work as planned.

PR #537, opened 2025-01-09
Apply hints to IO panels

liamhuber:
TODO: 
- [x] deal with `HasIO`
- [x] refactor/revise

### DIFF
--- a/pyiron_workflow/io.py
+++ b/pyiron_workflow/io.py
@@ -341,14 +341,14 @@ class HasIO(HasStateDisplay, HasLabel, HasRun, ABC):
         return self.inputs.connected or self.outputs.connected or self.signals.connected
 
     @property
-    def fully_connected(self):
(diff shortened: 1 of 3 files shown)


pyiron_workflow/io.py
```diff
@@ -52,7 +57,9 @@ class IO(HasStateDisplay, ABC):
     be assigned with a simple `=`.
     """
 
-    def __init__(self, *channels: Channel):
+    channel_dict: DotDict[str, OwnedType]
+
+    def __init__(self, *channels: OwnedType):
         self.__dict__["channel_dict"] = DotDict(
```
XzzX, line 63: Access via `__dict__`?

liamhuber: Because we override `__setattr__`. Instead of encoding a special exception case for a particular key there, I use dict to just avoid traveling that path.

Issue #360, opened 2024-06-10
Refactor: `Workflow` not a `Node`, and all `Node` with as `HasIOPreview`


liamhuber:
Right now `Workflow(ParentMost, Composite)` ultimately inherits from `Node` and `StaticNode(Node, HasIOPreview, ABC)` has class-level IO but is a child of `Node`.

I would like to refactor so that `Workflow` does _not_ inherit from `Node`. In this way, everything that's actually a "node" has class-level IO, and that can be a true feature of "nodes". This feature is important for things like node recommendation systems, and just generally a really useful thing to have. `Workflow` meanwhile is a lot _like_ a `Node` (they probably need a shared ancestor), but is not one itself -- it always needs to be parent-most in it's graph, and has dynamic IO.

Then we can remove the creator instance from `Composite` and _just_ have it on `Workflow`. At that point it should be possible to avoid all the delayed imports we have on `Creator` (and similar interface classes) and just leverage things directly. This would be great because the current delayed imports really screw up propagating the docstrings! (I'm not 100% sure yet if the reparenting will be sufficient to avoid circular imports or if a few more adjustments will be needed, but I think it will at least be possible.) The interfaces could then eventually be massaged into a formal public API to guide versioning decisions.

XzzX:
Is it intended to modify the workflow after its initial creation?

liamhuber:
They are intended to have both flexible child nodes and thus also -- critically -- flexible IO.

For macros, we do allow modification of the child nodes (ie subgraph) -- which I don't like and would rather stop if we can -- but the IO pattern is totally fixed at the class level.

liamhuber:
Pre-`flowrep` issue

PR #538, opened 2025-01-10
`mypy` semantics

liamhuber:
Resolves most of the mypy complaints in the `semantics` module. The two main changes are to make `SemanticParent` a generic, and purging the `ParentMost`. 

For the former, the reliance on `Generic` allows us to specify a type for the children, which is leveraged in `Composite(SemanticParent[Node],...)`. In the long run, I believe this should be narrowed to `SemanticParent[StaticNode]`, but that's related to #360 and so I want to hold off for now.

For the latter, the only place this was really relevant was in `Workflow`. It was anyhow not a great solution, since in principle we _do_ want `Workflow` to be able to have semantic parents...just not workflow graph parents. In the absence of `ParentMost`, it is simply up to subclasses of `Semantic` to update their `parent` setter if they want restrictions on their parentage beyond the standard `SemanticParent | None`. IMO this is both cleaner and a step toward the planned idea of non-workflow semantic parents for organizing groups of workflows.

Not dealt with here are some complaints about the `label` argument in `super` calls. The tension I'm running into is that for many nodes, the label can be perfectly well inferred from the class name and/or the node's location in a graph, and so a keyword is fine; while for `Workflow` it is always necessary to provide a name since this is the opening semantic scope and we don't want to start _every_ new workflow with `"Workflow/..."` -- _but_ it would be much nicer to write `Workflow("my_name")` than `Workflow(label="my_name")`. I just want more time to think about the most convenient solution.

### DIFF
--- a/pyiron_workflow/mixin/semantics.py
+++ b/pyiron_workflow/mixin/semantics.py
@@ -63,6 +63,13 @@ class Semantic(UsesState, HasLabel, HasParent, ABC):
 
     @parent.setter
     def parent(self, new_parent: SemanticParent | None) -> None:
+        self._set_parent(new_parent)
+
+    def _set_parent(self, new_parent: SemanticParent | None):
+        """
+        mypy is uncooperative with super calls for setters, so we pull the behaviour
+        out.
+        """
         if new_parent is self._parent:
             # Exit early if nothing is changing
             return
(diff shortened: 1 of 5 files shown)


liamhuber:
There are also two errors that I'm not getting locally, even though I've verified that my local machine and the CI are both running 1.14.1. Looking at the code, I agree with my machine -- it looks fine. Of course we'll need to get them resolved on the CI eventually,  but for now I want to push on working on something substantive.

```
pyiron_workflow/mixin/semantics.py:363: error: Item "str" of "ChildType | str" has no attribute "parent"  [union-attr]
pyiron_workflow/mixin/semantics.py:365: error: Incompatible return value type (got "ChildType | str", expected "ChildType")  [return-value]
```

PR #540, opened 2025-01-10
Refactor connection validity

liamhuber:
The instance check to see if a connection candidate has the correct (conjugate) type now occurs only _once_ in the parent `Channel` class. `Channel._valid_connection` is the repurposed to check for validity inside the scope of the classes already lining up, and defaults to simply returning `True` in the base class. `DataChannel` overrides it to do the type hint comparison.

Changes inspired by [conversation](https://github.com/pyiron/pyiron_workflow/pull/533#discussion_r1908526844) with @XzzX.


XzzX:
I like it!

PR #542, opened 2025-01-10
`mypy` topology and find

liamhuber:
Nothing of great interest here.


liamhuber:
Ahh, oops, I forgot to check out a separate branch. IMO it's not worth the git of fixing, this was all quite benign stuff.

PR #543, opened 2025-01-14
fix unnecessary explicit instantiations

XzzX:
Many functions can directly take generators. We do not need to explicitly instantiate lists, etc before passing them to the functions. Also use generator expressions where possible.

Enables and fixes ruff `C4` checks.


liamhuber (approved):
Super, LGTM 🚀

PR #544, opened 2025-01-15
Semantics generic parent

liamhuber:
Some refactoring, namely getting rid of the `HasParent` interface (which was only used in a single place, `Semantic`, which can simply have that as part of _its_ interface) and breaking `SemanticParent` out of the `Semantic` inheritance tree. These made it easier to then make `Semantic` generic on its parent type, so that type narrowing in `Node(..., Semantic["Composite"], ...)` gives us helpful type narrowing.

I think `mypy --strict` will still complain about the typevar definitions,

```python
ParentType = TypeVar("ParentType", bound="SemanticParent")
ChildType = TypeVar("ChildType", bound="Semantic")
```

This is because the bounds are generic and we leave the resolution implicitly as `typing.Any`. At present I'm not concerned about this because (a) let's get `mypy` happy before worrying about `mypy --strict`, and (b) because of the circular way that `Semantic(Generic[ParentType])` and `SemanticParent(Generic[ChildType])`, I think we are anyhow actually achieving 

```python
ParentType = TypeVar("ParentType", bound="SemanticParent[Semantic[SemanticParent[...]]")
ChildType = TypeVar("ChildType", bound="Semantic[SemanticParent[Semantic[...]]")
```

even though we obviously can't actually write that out.

TODO: Take a swing at delaying the specification of `Semantic`'s generic type in `Node`, such that we don't need explicit handling of the parent-most behaviour of `Workflow`, but can do something like `Workflow[Node[None]]`, while `StaticNode[Node[Composite]]` continues to behave in the current way. (It might be necessary to delay this until #360 is resolved and we can directly say `Workflow[Semantic[None]]` and `Node[Semantic[Composite]]` (or similar).

### DIFF
--- a/pyiron_workflow/mixin/has_interface_mixins.py
+++ b/pyiron_workflow/mixin/has_interface_mixins.py
@@ -53,17 +53,6 @@ class HasLabel(ABC):
         return self.label
 
 
-class HasParent(ABC):
-    """
-    A mixin to guarantee the parent interface exists.
-    """
-
-    @property
-    @abstractmethod
-    def parent(self) -> Any:
-        """A parent for the object."""
-
-
 class HasChannel(ABC):
     """
     A mix-in class for use with the :class:`Channel` class.
--- a/pyiron_workflow/mixin/has_interface_mixins.py
+++ b/pyiron_workflow/mixin/has_interface_mixins.py
@@ -11,7 +11,7 @@ possible coupling between different components of a composed class.
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING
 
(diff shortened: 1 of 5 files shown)


liamhuber:
> TODO: Take a swing at delaying the specification of `Semantic`'s generic type in `Node`, such that we don't need explicit handling of the parent-most behaviour of `Workflow`, but can do something like `Workflow[Node[None]]`, while `StaticNode[Node[Composite]]` continues to behave in the current way. (It might be necessary to delay this until #360 is resolved and we can directly say `Workflow[Semantic[None]]` and `Node[Semantic[Composite]]` (or similar).

`None` is anyhow outside the bounds of the typevar for semantic parents, so indeed let's postpone this change and hope it's even easier later.

PR #545, opened 2025-01-15
avoid commented-out code (ruff-ERA)

XzzX:
Some of the pieces are commented well, however, since we cannot quarantee good comments I suggest avoiding commented-out code entirely to avoid confusion.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,6 +84,8 @@ select = [
     "I",
     # flake8-comprehensions
     "C4",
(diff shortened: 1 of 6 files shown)


liamhuber (approved):
Yeah, sounds good to me. Where the commented code was intentional to highlight something, I just replaced it with an equivalent english statement. Otherwise I went in and resolved the remaining complains by a combination of deletion and uncommenting

PR #550, opened 2025-01-16
Make the `HasChannel` interface generic on the `Channel` type

liamhuber:


liamhuber:
In my first attack I had made `HasChannel` generic directly, and then tried to define `class Channel(HasChannel[Self], ...)`. This is beautiful, but `Self` isn't yet available at that point, so mypy complains anyhow. Attacks in this direction with a `SelfType` analogous to `ConjugateType` wound up being super verbose and ultimately less specific. In all the applications of `HasChannel` we really just care that it has a channel, so I don't presently see a performance issue by letting `Channel(HasChannel...)` and `ExploitsSingleOutput(..., HasGenericChannel[OutputDataWithInjection], ...)` each narrow the hint on the property in their own way. I'm open to revisiting again later if necessary.

XzzX (approved):

PR #568, opened 2025-01-21
Make Macro.graph_creator a normal method

liamhuber:
With accompanying change to applications/invocations, type hint narrowing, and docstrings


XzzX (approved):
I cannot

Issue #566, opened 2025-01-20
`value_receiver` vs `_value_receiver`?


XzzX:
```python
class DataChannel(FlavorChannel["DataChannel"], typing.Generic[ReceiverType], ABC):
    def __init__(
        self,
        label: str,
        owner: HasIO,
        default: typing.Any | None = NOT_DATA,
        type_hint: typing.Any | None = None,
        strict_hints: bool = True,
        value_receiver: ReceiverType | None = None,
    ):
        ...
        self.value_receiver: ReceiverType = value_receiver

    @property
    def value_receiver(self) -> Self | None:
        return self._value_receiver
```

Is this a bug, or is this intended? `value_receiver` vs `_value_receiver`?

liamhuber:
This is completely intended. `value_receiver` is a setter/getter property. The underlying `_value_receiver` is earlier set to some default value, and this pattern is leveraged to have the setter sanitize whatever `value_receiver` arg the user has passed into `__init__`

``python
    def __init__(
        ...
        value_receiver: ReceiverType | None = None,
    ):
        ...
        self._value_receiver = None
        ....
        self.value_receiver: ReceiverType = value_receiver
    
    ...

    @value_receiver.setter
    def value_receiver(self, new_partner: Self | None):
        if new_partner is not None:
            ...
```

**Comment by XzzX** [comment]
```python
self.value_receiver: ReceiverType = value_receiver
```
The type hint is misleading as it suggests a declaration shadowing your property.  
You could also get rid of `ReceiverType` entirely as you only allow `Self`.

liamhuber:
You're right that the attribute doesn't need a hint since the property has one, but the property hint should actually be `ReceiverType` not `Self`. We can't get rid of `ReceiverType` because there is a subclass of the output data channel: https://github.com/pyiron/pyiron_workflow/pull/569

XzzX:
> You're right that the attribute doesn't need a hint since the property has one, but the property hint should actually be `ReceiverType` not `Self`. We can't get rid of `ReceiverType` because there is a subclass of the output data channel: [#569](https://github.com/pyiron/pyiron_workflow/pull/569)

Fallen into my own trap... I am still confused by the meaning of Channel. But yes, of course, you are right.

PR #569, opened 2025-01-21
Fix value receiver hinting

liamhuber:
No need to hint the property, and because an `OutputDataWithInjection` can take an `OutputData` as a value receiver, the hint on the property should be the receiver type, not self.

### DIFF
--- a/pyiron_workflow/channels.py
+++ b/pyiron_workflow/channels.py
@@ -410,7 +410,7 @@ class DataChannel(FlavorChannel["DataChannel"], typing.Generic[ReceiverType], AB
         return self._value_receiver
 
     @value_receiver.setter
-    def value_receiver(self, new_partner: Self | None):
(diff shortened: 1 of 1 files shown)


XzzX (approved):

XzzX (changes requested):
https://github.com/pyiron/pyiron_workflow/pull/569/files#diff-fb0ac7f689d5a746d6b22d4902cc0607761f12e967538041ce16ee21b15bfa16R413

This should also be changed?

liamhuber:
Aha, link didn't work but I bet you're talking about the setter hint. Good catch, that was sloppy of me! I'll take another quick peek later this morning for any other missed spots and correct them

liamhuber:
IMO this now closes #566

XzzX (approved):

PR #570, opened 2025-01-23
Use inspect in order to get type hint metadata

samwaseda:
I was trying to get full type hints by using `inspect` instead of `get_type_hints`.

Well, this being said I'm painfully aware of the fact that the tests fail, but I don't really know why XD The problem comes from the fact that somehow `type_dict["return"]` turns the return type into a string, so for example instead of `tuple[float, float, float]` you get `"tuple[float, float, float]"`, even though if I copy and paste the code snippet I can get the correct type. Is there any transformation that takes place before `preview_io`?

### DIFF
--- a/pyiron_workflow/mixin/preview.py
+++ b/pyiron_workflow/mixin/preview.py
@@ -22,7 +22,6 @@ from typing import (
     ClassVar,
     get_args,
 )
-import inspect
 
 from pyiron_snippets.dotdict import DotDict
 
--- a/pyiron_workflow/mixin/preview.py
+++ b/pyiron_workflow/mixin/preview.py
@@ -200,12 +200,15 @@ class ScrapesIO(HasIOPreview, ABC):
         The result of :func:`inspect.signature` on the io-defining function
         """
         sig = inspect.signature(cls._io_defining_function(), eval_str=True)
-        type_dict = {key: value.annotation for key, value in sig.parameters.items()}
-        type_dict = type_dict | {"return": sig.return_annotation}
-        type_dict = {k: v for k, v in type_dict.items() if v != inspect.Parameter.empty}
-        for key, value in type_dict.items():
-            if value is None:
-                type_dict[key] = type(None)
+        type_dict = {
+            key: value.annotation if value.annotation else type(None)
+            for key, value in sig.parameters.items()
+            if value.annotation != inspect.Parameter.empty
+        }
+        if sig.return_annotation != inspect.Parameter.empty:
+            type_dict["return"] = (
+                sig.return_annotation if sig.return_annotation else type(None)
+            )
         return type_dict
 
     @classmethod
--- a/pyiron_workflow/mixin/preview.py
+++ b/pyiron_workflow/mixin/preview.py
@@ -199,7 +199,7 @@ class ScrapesIO(HasIOPreview, ABC):
         """
         The result of :func:`inspect.signature` on the io-defining function
         """
-        sig = inspect.signature(cls._io_defining_function())
+        sig = inspect.signature(cls._io_defining_function(), eval_str=True)
         type_dict = {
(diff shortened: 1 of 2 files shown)


liamhuber:
I don't think I'll have the time to support and dig deep on this until the end of next week, but to try to help a little at first glance: the error looks as though the code want's a tuple of hints (even if they're None) but the new routine provides an empty tuple. Since you're replacing one standard library call for a different one (plus some manipulation), I guess that for now it's just a matter of getting the new call + manipulation to conforming to the return type of the old one -- i.e. I don't expect complications from `pyiron_workflow`, at this stage it should all be totally local interactions with the standard library.

> Is there any transformation that takes place before `preview_io`?

No, what you're doing here _is_ the transformation. Subclasses of classes like `Function` and `Macro` get the io-defining function assigned to them, and here we directly parse it. From what I see here/recall, there is no `pyiron_workflow` magic here; per above you're just passing some function to the standard library inspection tools.

samwaseda:
> I don't think I'll have the time to support and dig deep on this until the end of next week

That's fine. I was anyway going to work on it myself. I just thought you might know where the problem comes from.

Maybe I make it a bit more specific what I wanted to say in my post above: The following snippet delivers the **same** results for a given function `f` as `get_type_hints(f)`:

```python
sig = inspect.signature(f)
type_dict = {
    key: value.annotation for key, value in sig.parameters.items()
}
type_dict = type_dict | {"return": sig.return_annotation}
type_dict = {
    k: v for k, v in type_dict.items() if v != inspect.Parameter.empty
}
```

The weird thing is, inside `pyiron_workflow` somehow `"return": sig.return_annotation` is turned into a string. So in my notebook, I get the error message about the following function:

```python
@as_function_node("started_at", "ended_at", "was_empty")
def ChangeDirectory(
    path: str | Path,
    delete_start: bool = False,
) -> tuple[str, str, bool]:
```

and `tuple[str, str, bool]` is turned into `"tuple[str, str, bool]"`, then by [applying `get_args`](https://github.com/pyiron/pyiron_workflow/blob/b3c028962c3bc6ffebbfed2a8c8c04d3b7b4874e/pyiron_workflow/mixin/preview.py#L165) it's turned into an empty tuple. So the symptoms can be explained. I just don't understand the cause.

Anyway I was thinking about working on it myself, so if nothing comes to your mind, just leave it like this.

samwaseda:
Thanks to [this post](https://github.com/marimo-team/marimo/issues/3312) I found out the reason - apparently `from __future__ import annotations` messes up the things. There was an argument `eval_str` that I had to set to `True`. Now it should be fine.

pyiron_workflow/mixin/preview.py
```diff
@@ -198,9 +197,16 @@ def _get_output_labels(cls):
...
+        The result of :func:`inspect.signature` on the io-defining function
         """
-        return get_type_hints(cls._io_defining_function())
+        sig = inspect.signature(cls._io_defining_function(), eval_str=True)
+        type_dict = {key: value.annotation for key, value in sig.parameters.items()}
+        type_dict = type_dict | {"return": sig.return_annotation}
+        type_dict = {k: v for k, v in type_dict.items() if v != inspect.Parameter.empty}
+        for key, value in type_dict.items():
+            if value is None:
+                type_dict[key] = type(None)
```
liamhuber, line 208: I like this better, but it's not a hill I'll die on.

```suggestion
        type_dict = {
            key: value.annotation if value.annotation else type(None)
            for key, value in sig.parameters.items()
            if value.annotation != inspect.Parameter.empty
        }
        if sig.return_annotation != inspect.Parameter.empty:
            type_dict["return"] = (
                sig.return_annotation if sig.return_annotation else type(None)
            )
```

liamhuber (approved):
LGTM

Optional refactoring suggestion.

PR #571, opened 2025-01-24
Refactor preview

samwaseda:
I also refactored it, because now everywhere we use `inspect.signature` and from my point of view there's not much of a reason to separate the functions.


liamhuber (approved):
LGTM

PR #572, opened 2025-01-24
Update tests

samwaseda:
I updated the type checking because otherwise it fails with annotated typing

### DIFF
--- a/tests/unit/test_type_hinting.py
+++ b/tests/unit/test_type_hinting.py
@@ -6,6 +6,7 @@ from pint import UnitRegistry
 from pyiron_workflow.type_hinting import (
     type_hint_is_as_or_more_specific_than,
     valid_value,
+    _get_type_hints,
 )
 
 
@@ -95,6 +96,16 @@ class TestTypeHinting(unittest.TestCase):
                 typing.Callable[[int, float], float],
                 False,
             ),
+            (
+                typing.Annotated[int, "foo"],
+                int,
+                True,
+            ),
+            (
+                int,
+                typing.Annotated[int, "foo"],
+                True,
+            ),
         ]:
             with self.subTest(
                 target=target, reference=reference, expected=is_more_specific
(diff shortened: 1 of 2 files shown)


liamhuber:
Sorry, I didn't manage to get to this today. Should get it tomorrow.

tests/unit/test_type_hinting.py
```diff
@@ -95,15 +96,37 @@ def test_hint_comparisons(self):
...
             ):
                 self.assertEqual(
                     type_hint_is_as_or_more_specific_than(target, reference),
                     is_more_specific,
+                    msg=f"{target} is {'not ' if not is_more_specific else ''}more specific than {reference}",
                 )
 
+    def test_get_type_hints(self):
+        for hint, origin in [
+            (int | float, type(int| float)),
```
liamhuber, line 121: Is it really intentional to do `type(int | float)` and not `type[int | float]` here?

tests/unit/test_type_hinting.py
```diff
@@ -95,15 +96,37 @@ def test_hint_comparisons(self):
...
+            (typing.Annotated[int | float, "foo"], type(int | float)),
+            (int, None),
+            (typing.Annotated[int, "foo"], None),
```
liamhuber, line 124: Could we extend this to include hints that specify generics? E.g.

```suggestion
            (typing.Annotated[int, "foo"], None),
            (typing.Annotated[list[int], list),
            (list[int], list),
```

liamhuber (commented):
Ok, if you read my commit suggestion and it makes sense and passes the tests, then I've understood this well enough that I've reviewed it and it's approved by me. 

But, if you look at that suggestion and you wonder what the heck I was thinking, then either you need to hold my hand or I need to check out the branch and try stuff out, because I've misunderstood. In that case my approval is worth what I've understood -- nothing 😝

tests/unit/test_type_hinting.py
samwaseda: I edited it because there was a typo, and `typing.Annotated` doesn't work without metadata.

tests/unit/test_type_hinting.py
samwaseda: What does `type[int | float]` mean?

liamhuber (approved):
Nice, looks like I understood ok. LGTM.

tests/unit/test_type_hinting.py
liamhuber: It is a type hint, that we expect the type `int` or the type `float`. In this case `bool` also works, since it's a subclass of `int`, but, e.g. `string` doesn't.

The `int | float` is generating a typing union, and we're treating `type` as a generic (i.e. takes subtyoes like `list[]`, so what we've written is shorthand for `x: type[int] | type[float]`.

Concretely, the following runs and mypyp will complain only about `foo(str)`:

```
def foo(x: type[int, float]) -> None:
    print(x)

foo(int)  # Fine -- in the type hint
foo(bool)  # Fine -- subclass of int
foo(str)  # mypy complains


def equivalent_to_foo(x: type[int] | type[float]) -> None:
    print(x)
```

`type(int | float)` is an instance of `types.UnionType`. Looking at it again, I guess this is actually exactly what you want, since the hint `int | float` should indeed have a `UnionType` origin. It's not clear to me whether it matters that you put `int | float` inside the union type, as I don't know how detailed its `==` comparison is.

PR #580, opened 2025-02-05
Fix comparisons tests

liamhuber:
Patching #579 

In particular, when `storage.available_backends` receives a `StorageInterface` instance it needs to return that instance _and_ the standard backend(s), because interfaces may have states so a simple type check is not sufficiently details. And when `storage.available_backends` receives a string, it should _not_ return two copies of the standard backend(s) because the string can't carry any state so the default standard backend must be identical.

Honestly, the whole thing could use more simplification, but I am again uninterested in perfecting it while a new storage infrastructure is currently being worked on. In the mean time, this on top of #579 anyhow moves things in a nice direction in the meantime.

### DIFF
--- a/pyiron_workflow/storage.py
+++ b/pyiron_workflow/storage.py
@@ -297,11 +297,14 @@ def available_backends(
     """
 
     standard_backends = {"pickle": PickleStorage}
-    backend = standard_backends.get(backend, PickleStorage)() if isinstance(backend, str) else backend
+    backend_instance = standard_backends.get(backend, PickleStorage)() if isinstance(backend, str) else backend
 
-    if backend is not None:
-        yield backend
+    if backend_instance is not None:
+        yield backend_instance
         if only_requested:
             return
 
-    yield from (v() for v in standard_backends.values() if not isinstance(backend, v))
+    yield from (
+        v() for k, v in standard_backends.items()
+        if (v != backend_instance and k != backend)
+    )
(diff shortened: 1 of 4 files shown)


liamhuber:
Ah no. I can't compare `backend != v` because I need `v()` instead.

Sorry, @XzzX I don't have any more time to sink into this issue today.

pyiron_workflow/storage.py
```diff
@@ -297,11 +297,15 @@ def available_backends(
...
+        else backend
+    )
+
+    if backend_instance is not None:
+        yield backend_instance
         if only_requested:
             return
 
-    yield from (v() for v in standard_backends.values() if not isinstance(backend, v))
+    yield from (v() for k, v in standard_backends.items() if k != backend)
```
XzzX, line 311: Well, you reverted it back to a semi-broken state. `k` is of type `str` and backend is of type `Literal["pickle"] | StorageInterface | None`. If you did not implement a custom `__eq__` operator this is broken for `StorageInterface`.

XzzX (changes requested):
I do not see the point of this function. It behaves differently for `str` input and `StorageInterface` input. Why would I put in a `StorageInterface` instance?

pyiron_workflow/storage.py
liamhuber: Sounds good to me, go ahead and stack it on

PR #582, opened 2025-02-07
Mypy tests too

liamhuber:
Demonstrate the resolution of "severe" mypy complaints in #575 by just declaring some `ClassVar`s

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install mypy
         run: pip install mypy
       - name: Test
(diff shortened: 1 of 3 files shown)


liamhuber:
@XzzX, you can see that both of [these mypy complaints](https://github.com/pyiron/pyiron_workflow/pull/575#issuecomment-2639360865) disappear as soon as we let it know that the class variables exist. These are really not big problems, just sloppiness in the test suite about fully indicating to the static type checker which class attributes it should expect to encounter.

XzzX:
-> #591

Issue #504, opened 2024-11-21
Architecture: can `Transformer` nodes be `Function`?


liamhuber:
In principle, the `Transformer` nodes all just run a simple function -- can we make them all inherit directly from `Function`? Of course the constructor calls to generate the transformer node classes dynamically will look different, since the point there is that we know basically what the function should look like but the number/names of I/O varies, whereas when we transform a function into a node the function signature is already fully known, but _after_ they're created they look remarkably similar. It would be nice to bring the architecture together this way to minimize the number of different node classes floating around -- ideally users should only ever have to deal with `Macro` and `Function`, where `Function` may come in the _flavour_ of transformer/dataclass nodes rather than those being something actually distinct.

XzzX:
I like the idea. Do you want to keep `For` as it is? If I get it correctly `For` is neither a `Macro` nor a `Function`.

liamhuber:
I like the idea of `For` being directly a `Macro`, but `For` is rather different in that it destroys and recreates (parts of) it child graph on each run. If we can nicely make it play as a child of `Macro`, I would be happy, I'm just not as overwhelming confident that it will work like I am for `Function` and `Transformer`.

As a first step, let's anyhow do this here, and get `While` working in the same paradigm as `For`, and then see about reparenting the loops.

liamhuber:
I should say, `For` subclasses _do_ have their IO specified at the class level, so I don't see a fundamental philosophical barrier to making them `Macro`s, at the moment I only have the lesser concern of technical challenge

liamhuber:
Pre-`flowrep` issue

PR #586, opened 2025-02-11
Export to dict

### LINKED ISSUE #576: `pyiron_workflow`-independent dict containing workflow info
Note: I will work on this myself. You can just tell me what you think @liamhuber 

Currently, we have `graph_as_dict`, which returns a dict containing all the information about the workflow. However, when we try to make `semantikon` work with any workflow manager, the most natural thing for me is to ask every workflow manager to provide a dictionary that does not require `semantikon` to depend on the workflow manager. In addition, that would be consistent with the effort that @jan-janssen is making [here](https://github.com/pyiron-dev/compare-workflow-graphs).

I was initially thinking we should extend @jan-janssen's universal representation of workflows by adding optional args required by `semantikon`, and potentially also allowing `nodes_dict` to contain sub-nodes to represent also the `Macro` nodes. However, I started realizing that this might not be the right way, because:

- There's no consensus on how macro nodes or metadata should be defined
- (Maybe more importantly) @jan-janssen's universal representation specifies the **numerical logic of execution** and in this regard it is conceptually complete

So my first suggestion is to separate `semantikon`-dict from the universal representation dict, in order to represent the workflow including the scientific logic.

Coming to the actual content, `semantikon` would need:

- IO-values
- Edges
- Unique node labels
- IO labels
- Underlying python function
- (Optionally) type hints

The type hints are optional, because `semantikon` can parse them, but that could be a bit dangerous because it might fail to assign the types to the IO labels. The reason why we need the original python function is because since [this PR](https://github.com/pyiron/semantikon/pull/43) the function metadata is attached to `f._semantikon_metadata`, which I would not ask the workflow manager to provide.

Concerning the `Macro` nodes, I was testing the following code:

```python
def add_one(a: int):
    result = a + 1
    return result

def add_two(b: int):
    result = b + 2
    return result

@Workflow.wrap.as_macro_node
def add_three(macro, c: int):
    macro.one = add_one(a=c)
    macro.two = add_two(b=macro.one)
    result = macro.two
    return result

wf = Workflow("my_wf")
wf.three = add_three(c=1)
wf.four = add_one(a=wf.three)
wf.run()
```

Then I realized that `wf.graph_as_dict` does explicitly not provide the connection between `wf.three.inputs.c` and `wf.three.one.inputs.a` as well `wf.three.two.outputs.result` and `wf.three.outputs.result`. They may not be edges in the sense of workflow edges, but they still somehow have to be included in the dict.

Well, thanks for reading, but actually I'm probably the one who profited from it the most as it helped me figure out what I actually want XD

samwaseda:
Closes #576 

I implemented what we have discussed in the issue above. Initially, I created an extra file called `export.py`, but then realized that I cannot import `Composite` there if I want to be able to export it from the workflow, so I simply put it inside `composite.py`. I was not so sure about the keys - you see a combination of `label`, `scoped_label` etc., but maybe there's a smarter way to tackle the problem.

Anyway for this workflow:

```python
def add_one(a: int):
    result = a + 1
    return result

def add_two(b: int = 10) -> int:
    result = b + 2
    return result

@Workflow.wrap.as_macro_node
def add_three(macro, c: int) -> int:
    macro.one = add_one(a=c)
    macro.two = add_two(b=macro.one)
    w = macro.two
    return w

wf = Workflow("my_wf")

wf.three = add_three(c=1)
wf.four = add_one(a=wf.three)

wf.run()

wf.export_to_dict()
```

You can get the dict:

```python
{'inputs': {'three__c': {'value': 1, 'type_hint': int}},
 'outputs': {'four__result': {'value': 5}},
 'nodes': {'three': {'inputs': {'three__c': {'value': 1, 'type_hint': int}},
   'outputs': {'three__w': {'value': 4, 'type_hint': int}},
   'nodes': {'one': {'inputs': {'a': {'value': 1, 'type_hint': int}},
     'outputs': {'result': {'value': 2}},
     'function': <function __main__.add_one(a: int)>},
    'two': {'inputs': {'b': {'default': 10, 'value': 2, 'type_hint': int}},
     'outputs': {'result': {'value': 4, 'type_hint': int}},
     'function': <function __main__.add_two(b: int = 10) -> int>}},
   'edges': [('inputs.three__c', 'one.inputs.a'),
    ('one.outputs.result', 'two.inputs.b'),
    ('two.outputs.result', 'outputs.three__w')]},
  'four': {'inputs': {'a': {'value': 4, 'type_hint': int}},
   'outputs': {'result': {'value': 5}},
   'function': <function __main__.add_one(a: int)>}},
 'edges': [('inputs.three__c', 'one.inputs.a'),
  ('three.outputs.w', 'four.inputs.a')]}
```


samwaseda:
I'm not quite done with the unit tests, but I wanted to have feedback from @liamhuber before I move things around.

samwaseda:
So, `mypy` is failing and I kind of understand why - I guess it wants to have the same type for all the `dict` entries, or otherwise at least explicitly have the hint that says what kind of types are expected. I don’t really know how to make it compatible with what `mypy` expects - @liamhuber @XzzX ?

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,118 @@
...
+        data_dict.pop("value")
+    if not with_default:
+        data_dict.pop("default")
+    for key, value in data_dict.items():
+        if getattr(item, key) is not value:
+            data[key] = getattr(item, key)
+    return data
+
+
+def _is_internal_connection(channel: Channel, workflow: Composite, io_: str) -> bool:
```
liamhuber, line 43: \*sigh\*. What I would like is _all_ channels are _only ever_ connected to other channels in the same scope. Unfortunately, nodes with no parent at all at least complicate this. Anyhow, this function isn't bad, it's just that it _ought_ to be sufficient to merely use `channel.connected`. A cleaner solution would be to first go close #587, which I just opened because here you reminded me that we have this problem 😂 then come back here and not need this function at all.

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+                    inp, with_values=with_values, with_default=with_default
+                )
+            else:
+                data[io_][inp.label] = _extract_data(
+                    inp, with_values=with_values, with_default=with_default
+                )
+    return data
+
+
+def export_node_to_dict(
```
liamhuber, line 82: This will not work. You hint `Node`, but not all `Node` have `node_function`. Hinting `Function` would be too narrowly scoped, as we do have this possibility for stuff like `Transformer`.

It is also not a good feeling to have `export_node_to_dict` in `composite.py`. The scoping is off. I guess we want to be able to convert nodes to dicts regardless of whether they're composite or not! So I think such tools need to either exist locally next to the thing they're converting, or be collected together in some third location like `export.py` as you mentioned elsewhere.

I also think that the composite's probably should also get a "node function", but it's something like `Composite._on_run`, or any other placeholder for the fact that we're going to execute the subgraph and how we'll do that.

Honestly, I think you will have a much easier time with this whole PR if you (or you wait for me eventually to) clean up the space you're working in -- i.e. if #504 is closed, we'll have a much cleaner division of "composite" vs "atomic" nodes right in the class hierarchy. #360 might help too, but it's harder and I think you can get away without it here.

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+
+    Args:
+        workflow (Composite): The composite to export.
+        with_values (bool): Whether to include the values of the channels in the
+            dictionary. (Default is True.)
+
+    Returns:
+        dict: The exported composite as a dictionary.
+    """
+    data = {"inputs": {}, "outputs": {}, "nodes": {}, "edges": []}
```
liamhuber, line 115: All composite's are `Node`s, yet we don't use `export_node_to_dict` as a starting point -- this makes me uncomfortable. An immediate result of this is that we wind up with the `data` dict being re-declared here and we're just relying on future maintainers to make sure all the string keys stay nicely synchronized.

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+
+    Returns:
+        bool: Whether the channel is connected to another channel in the same workflow.
+    """
+    if not channel.connected:
+        return False
+    return any(channel.connections[0] in getattr(n, io_) for n in workflow)
+
+
+def _get_scoped_label(channel: Channel, io_: str) -> str:
```
liamhuber, line 60:
```suggestion
def _scoped_label_to_io_label(channel: Channel, io_: str) -> str:
```

Function name lies!

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+    data = {"inputs": {}, "outputs": {}, "nodes": {}, "edges": []}
+    for inp in workflow.inputs:
+        if inp.value_receiver is not None:
+            data["edges"].append(
+                (
+                    f"inputs.{inp.scoped_label}",
+                    _get_scoped_label(inp.value_receiver, "inputs"),
+                )
+            )
+    for node in workflow:
```
liamhuber, line 124: This is 100% correct, but it reminds me that I meant to open another issue... https://github.com/pyiron/pyiron_workflow/issues/588

We could proceed with such a definition now, but we'd need to remember to update it as part of that issue.

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+        if isinstance(node, Composite):
+            data["nodes"][label] = export_composite_to_dict(
+                node, with_values=with_values
+            )
+        else:
+            data["nodes"][label] = export_node_to_dict(node, with_values=with_values)
+        for inp in node.inputs:
+            if _is_internal_connection(inp, workflow, "outputs"):
+                data["edges"].append(
+                    (
```
liamhuber, line 135: I still advocate for `{}` rather than `()`

pyiron_workflow/nodes/composite.py
```diff
@@ -26,6 +27,130 @@
...
+                    )
+                )
+        for out in node.outputs:
+            if out.value_receiver is not None:
+                data["edges"].append(
+                    (
+                        _get_scoped_label(out, "outputs"),
+                        f"outputs.{out.value_receiver.scoped_label}",
+                    )
+                )
```
liamhuber, line 147: This is super similar to the same scraping for the input value receiver pairs, just a sort of conjugate. I think it should be possible to nicely pull out a `_get_value_synchronization_edges(sender: Node, receiver: Node, io: str)` function

liamhuber (changes requested):
Top most feedback: cool, good momentum!

Then: we absolutely can't merge such a thing as long as it is unsafely bound to `Function` nodes. This needs to work universally or not at all.

Finally: I think it is going to be possible to go in and add more branching clauses such that you safely cover our entire space of available nodes. However, in the spirit of "don't do hard things", IMO a better option is to clean up the stuff in `pyiron_workflow` that is making this a pain in the butt, and _then_ come implement the dictionary abstraction. I linked some existing issues, and in the process of the review wrote two new ones. If we get "Then:" resolved, I'm open to merging, I just think it's faster at the cost of more net work and more net trouble.

liamhuber:
> So, `mypy` is failing and I kind of understand why - I guess it wants to have the same type for all the `dict` entries, or otherwise at least explicitly have the hint that says what kind of types are expected. I don’t really know how to make it compatible with what `mypy` expects - @liamhuber @XzzX ?

I haven't dug into the error reports, but given the nature of the problem I suspect you're going to need something akin to `NestedDictAlias: TypeAlias = dict[str, "str | object | NestedDictAlias"]`

PR #589, opened 2025-02-12
Add executorlib container

### LINKED ISSUE #584: Upgrade `create` interface to be compliant with `executorlib` 0.1.0
#583 is going to fail because we `from executorlib import Executor`, which is no longer available since this was decomposed into a variety of more specific executors.

I see three likely resolutions:
1) The most direct solution is to simply swap out our import for one of these more specific classes. If we choose the right one, I suspect this will work out of the box.
2) A slightly more elegant solution is to make a new container that imports all of the executors and gives access to them; i.e. instead of `Workflow.create.ExecutorlibExecutor`, we'd have something like `Workflow.create.executorlib.SingleNodeExecutor` or `Workflow.create.executorlib.SlurmJobExecutor`.
3) The final route would be to remove the creator-access to these altogether. I still want to test against `executorlib` in the test suite, and probably the notebooks where we sell it, but then they wouldn't need to be in the core dependencies.

I like both (2) and (3). Since the mandatory dependencies are so light for `executorlib`, I would probably go for (2) as a first attack.

I don't mind this issue being on my plate, but I'm not likely to be able to get to it imminently and it should be a relatively straightforward fix, so I laid out the attack patterns here in case someone else wants to take a crack at it in the meantime.

niklassiemer:
closes #584

### DIFF
--- a/pyiron_workflow/create.py
+++ b/pyiron_workflow/create.py
@@ -40,9 +40,7 @@ class Creator:
 
         self.function_node = function_node
 
-    @property
-    def executorlib(self):
-        return ExecutorlibExecutor()
+    executorlib = executorlib
 
     @property
     def standard(self):
--- a/pyiron_workflow/create.py
+++ b/pyiron_workflow/create.py
@@ -6,11 +6,20 @@ from __future__ import annotations
 
 from concurrent.futures import ProcessPoolExecutor, ThreadPoolExecutor
 
-from executorlib import Executor as ExecutorlibExecutor
+import executorlib
 from pyiron_snippets.dotdict import DotDict
 
 from pyiron_workflow.executors import CloudpickleProcessPoolExecutor
 from pyiron_workflow.nodes.function import as_function_node, function_node
+
+
+class ExecutorlibExecutor:
+    def __init__(self):
+        self.SlurmJobExecutor = executorlib.SlurmJobExecutor
+        self.SingleNodeExecutor = executorlib.SingleNodeExecutor
+        self.FluxJobExecutor = executorlib.FluxJobExecutor
+        self.FluxClusterExecutor = executorlib.FluxClusterExecutor
+        self.SlurmClusterExecutor = executorlib.SlurmClusterExecutor
 
 
 class Creator:
@@ -28,10 +37,12 @@ class Creator:
         self.ThreadPoolExecutor = ThreadPoolExecutor
         # Local cloudpickler
         self.CloudpickleProcessPoolExecutor = CloudpickleProcessPoolExecutor
-        # executorlib
-        self.ExecutorlibExecutor = ExecutorlibExecutor
 
         self.function_node = function_node
+
+    @property
+    def executorlib(self):
+        return ExecutorlibExecutor()
 
     @property
     def standard(self):
--- a/pyiron_workflow/create.py
+++ b/pyiron_workflow/create.py
@@ -13,15 +13,6 @@ from pyiron_workflow.executors import CloudpickleProcessPoolExecutor
 from pyiron_workflow.nodes.function import as_function_node, function_node
 
 
-class ExecutorlibExecutor:
-    def __init__(self):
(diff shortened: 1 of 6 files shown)


pyiron_workflow/create.py
```diff
@@ -6,13 +6,22 @@
...
 from pyiron_workflow.nodes.function import as_function_node, function_node
 
 
+class ExecutorlibExecutor:
+    def __init__(self):
+        self.SlurmJobExecutor = executorlib.SlurmJobExecutor
+        self.SingleNodeExecutor = executorlib.SingleNodeExecutor
+        self.FluxJobExecutor = executorlib.FluxJobExecutor
+        self.FluxClusterExecutor = executorlib.FluxClusterExecutor
+        self.SlurmClusterExecutor = executorlib.SlurmClusterExecutor
```
liamhuber, line 22:
```suggestion
    SlurmJobExecutor = executorlib.SlurmJobExecutor
    SingleNodeExecutor = executorlib.SingleNodeExecutor
    FluxJobExecutor = executorlib.FluxJobExecutor
    FluxClusterExecutor = executorlib.FluxClusterExecutor
    SlurmClusterExecutor = executorlib.SlurmClusterExecutor
```

I guess these can just be class attributes, no need to wait for the instance

pyiron_workflow/create.py
```diff
@@ -28,11 +37,13 @@ def __init__(self):
...
         # Local cloudpickler
         self.CloudpickleProcessPoolExecutor = CloudpickleProcessPoolExecutor
-        # executorlib
-        self.ExecutorlibExecutor = ExecutorlibExecutor
 
         self.function_node = function_node
 
+    @property
+    def executorlib(self):
+        return ExecutorlibExecutor()
```
liamhuber, line 45:
```suggestion
    executorlib = executorlib
```

Alternatively, we could simply directly expose the executorlib module API on the creator!

Cons: a little less tightly controlled
Pros: insanely succinct

Since @jan-janssen is following semantic versioning and exposing the key API in the top-level `__init__`, I think I lean towards this solution

liamhuber:
@niklassiemer, I guess this

```
ERROR: test_executors (integration.test_workflow.TestWorkflow.test_executors)
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/integration/test_workflow.py", line 152, in test_executors
    Workflow.create.ExecutorlibExecutor,
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'Creator' object has no attribute 'ExecutorlibExecutor'

```

Just needs to be replaced with the relevant `Workflow.create.excutorlib...` counterpart, then we're good to go.

I suggest adopting one of the two suggestions I made to tighten up the interface, but if you hate both I can live with it as-is and otherwise once the tests are passing I'm happy here. Thanks for taking care of it!

niklassiemer:
Thanks for the feedback? I was actually using the tests here to figure out what needs to be changed in the tests xD and I was thinking to add `__all__` Executors in a loop. However, I like the class attributes more :)

Indeed, I would like to have `pyiron_workflow` aligned to executorlib asap to do a new release and allow the cmti environment to use the newest base / atomistics/workflow stack.

niklassiemer:
Funny enough all my code will be gone and you did the full work after all xD

niklassiemer:
I should not try these things on mobile xD do we want to maintain downwards compatibility to older versions of executorlib? Otherwise I would need to target the executorlib update branch I suppose :)

niklassiemer:
@liamhuber I do not know why the coveralls/codacy is failing - is it related to some hickups (I saw a comment of Jan on another PR but I do not remember where and I am not sure if this is related).
From my perspective this is good to go otherwise. Also the API change is rather small (`....create.ExecutorlibExecutor` -> `....create.executorlib.Executor` or `....create.executorlib.SingleNodeExecutor`; depending on the version of executorlib).
If you are happy, I would like a merge and a timely new release, such that I can get this into the cmti environment on Monday.

liamhuber (approved):

PR #590, opened 2025-02-13
enable nbqa in ci

XzzX:


XzzX:
```python
from typing_extensions import deprecated

@deprecated("use new_function")
def old_function() -> None:
    print("I am old")
```
works for all classes and functions defined in python files.  
Dynamically generated stuff (in our case most probably everything created via decorators) is not checked as it is only defined at runtime. I think
`warnings.warn(message, DeprecationWarning, stacklevel=2)`
is the way to go there. With the correct filter warnings can be converted to errors.

Starting with python>=3.13 also the `deprecated` decorator from warning can be used.

liamhuber:
I'm on mobile so it's inconvenient to double check, but I'm 90% confident the `dataclass` field is already there and properly stated on the `Dataclass` node parent class. I think going back and fixing up the typing on the decorators here and the ’pyiron_snippets.factory` code is likely to resolve this (and other) issues. For the next couple weeks my focus needs to be on the storage tools and pyironflow web portal though, so at least from my end there won't be any imminent action here. It's still somewhere on the todo list.

XzzX:
Ok, let's put this on hold for now.

PR #591, opened 2025-02-13
Enable mypy for tests

XzzX:


XzzX:
Hmm, perhaps `enable_error_code = []` silenced more than I thought.

PR #596, opened 2025-02-17
Out-of-process caching

liamhuber:
Here is a draft of out-of-process caching.

At the moment, it doesn't actually leverage a database, but uses the filesystem as a database -- a first-attack suggested by @XzzX. In this case, we can provide a file cache directory to write to/read from for any `StaticNode` (i.e., presently anything but a `Workflow`). When this argument is provided, the node will generate its hash using `pyiron_database.instance_database.node.get_hash`, and see if there's a file of the same name. If so, it uses `bagofholding` to load that data to its output and short-circuits its run routine in the same manner as the old in-process cache. Similarly, at the end of its run, it uses `bagofholding` to save its `self.outputs.to_value_dict()` to a file based on the `pyiron_database` hash.

Here I'm not concerned about minor inefficiencies like calculating the hash multiple times, or even re-saving data that has already been saved. I'm only concerned with getting it running and stable. Stability demands tests, but since neither `bagofholding` nor `pyiron_database` are yet available as dependencies, I didn't make any effort here yet. I'm just trying to get the ball rolling.

Usage:

```python
import time

import pyiron_workflow as pwf

@pwf.as_function_node
def sleepy(duration: int):
    time.sleep(duration)
    return duration

wf = pwf.Workflow("my_file_caching")
wf.s = sleepy(5, file_cache=".")
wf()  # 5s 23ms
```

Next cell

```python
wf2 = pwf.Workflow("child_gets_the_same_hash")
wf2.t = sleepy(5, file_cache=".")
wf2()  # 10ms
```

That is obviously a super primitive hash, since it directly points to an integer. But, when I applied it to [@samwaseda's example workflow](https://github.com/pyiron/storage/discussions/2#discussioncomment-12185002) setting the `file_cache` on the `wf.bulk_energy` and `wf.vac_energy` nodes it also worked splendidly.

Since `get_hash` already snags the node module+qualname and versioning (if available), I think we've got a good safety net for making sure only appropriate saves get loaded. `bagofholding` doesn't actually _do_ anything with versioning data yet, but in principle once it does, when the output is some complex data type, `bagofholding` will take care of making sure that the current env is appropriate for reloading it. At the end of the day, we may wish to expose some version bound kwargs in `pyiron_workflow` as they come to exist in `pyiron_database` and `bagofholding`, but I don't think `pyiron_workflow` itself needs to manage any versioning here 🚀 

Moving forward:
- Do we want to provide an `InstanceDatabase` instead of a directory (or directories, for reading)?
  - Do we need _both_, or will the `InstanceDatabase` know about where to move the files to?
- What is the release plan for `pyiron_database` so we can include it as a dependency?
- Releasing `bagofholding` still waits on the copyright ping from KIT

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -514,6 +514,7 @@ class Node(
         run_parent_trees_too: bool,
         fetch_input: bool,
         emit_ran_signal: bool,
+        **kwargs: Any,
     ) -> tuple[bool, Any]:
         if run_data_tree:
             self.run_data_tree(run_parent_trees_too=run_parent_trees_too)
@@ -522,19 +523,22 @@ class Node(
             self.inputs.fetch()
 
         if self.use_cache and self.cache_hit:  # Read and use cache
-            if self.parent is None and emit_ran_signal:
-                self.emit()
-            elif self.parent is not None:
-                self.parent.register_child_starting(self)
-                self.parent.register_child_finished(self)
-                if emit_ran_signal:
-                    self.parent.register_child_emitting(self)
-
-            return True, self._outputs_to_run_return()
+            return self._return_existing_result(emit_ran_signal)
         elif self.use_cache:  # Write cache and continue
             self._cached_inputs = self.inputs.to_value_dict()
 
         return super()._before_run(check_readiness=check_readiness)
+
+    def _return_existing_result(self, emit_ran_signal: bool) -> tuple[bool, Any]:
+        if self.parent is None and emit_ran_signal:
+            self.emit()
+        elif self.parent is not None:
+            self.parent.register_child_starting(self)
+            self.parent.register_child_finished(self)
+            if emit_ran_signal:
+                self.parent.register_child_emitting(self)
+
+        return True, self._outputs_to_run_return()
 
     def _run(
         self,
--- a/pyiron_workflow/nodes/for_loop.py
+++ b/pyiron_workflow/nodes/for_loop.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import itertools
 import math
+import pathlib
 from abc import ABC
 from concurrent.futures import Executor
 from functools import lru_cache
@@ -202,6 +203,7 @@ class For(Composite, StaticNode, ABC):
         delete_existing_savefiles: bool = False,
         autorun: bool = False,
         checkpoint: Literal["pickle"] | StorageInterface | None = None,
+        file_cache: str | pathlib.Path | None = None,
         strict_naming: bool = True,
         body_node_executor: Executor | None = None,
         **kwargs,
@@ -214,6 +216,7 @@ class For(Composite, StaticNode, ABC):
             autorun=autorun,
             autoload=autoload,
             checkpoint=checkpoint,
+            file_cache=file_cache,
             strict_naming=strict_naming,
             **kwargs,
         )
--- a/pyiron_workflow/nodes/static_io.py
+++ b/pyiron_workflow/nodes/static_io.py
@@ -2,8 +2,9 @@ from __future__ import annotations
 
 import pathlib
 from abc import ABC
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Any, Literal
 
+import bagofholding as boh
 from pandas import DataFrame
 from pyiron_snippets.colors import SeabornColors
 
@@ -17,6 +18,8 @@ from pyiron_workflow.mixin.preview import HasIOPreview
 from pyiron_workflow.node import Node
 
 if TYPE_CHECKING:
+    from concurrent import futures
+
     from pyiron_workflow.nodes.composite import Composite
     from pyiron_workflow.storage import StorageInterface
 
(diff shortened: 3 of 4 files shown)


liamhuber:
@jan-janssen and anyone else involved in the DPG presentation -- this is obviously still super preliminary, but feel free to poke your head in at any time and give feedback on the UI/UX. What value goes in the kwarg is very much unclear to me, but at the end of the day I do currently envision triggering out-of-process caching on a node-by-node basis with such an instantiation kwarg as above.

XzzX:
> `pyiron_database.instance_database.node.get_hash`

Please use `pyiron_database.instance_database.get_hash`. I do not like the node in there and will rename it at some point. Single point of entrence for the instance_database is `pyiron_database.instance_database`



> * Do we want to provide an `InstanceDatabase` instead of a directory (or directories, for reading)?

Perhaps to reduce the number of file. We can talk about this on Wednesday. We at some point need to think about the inode limit of the filesystem.

>   * Do we need _both_, or will the `InstanceDatabase` know about where to move the files to?

Single source of truth would be preferable.

> * What is the release plan for `pyiron_database` so we can include it as a dependency?

https://pypi.org/project/pyiron-database/  
Currently, I handle postgresql and neo4j as optional dependencies. Seemingly conda is not able to do this.

liamhuber:
> > `pyiron_database.instance_database.node.get_hash`
> 
> Please use `pyiron_database.instance_database.get_hash`. I do not like the node in there and will rename it at some point. Single point of entrence for the instance_database is `pyiron_database.instance_database`

Oops, sorry, missed the submodule `__init__` API. Done 👍 

> > * Do we want to provide an `InstanceDatabase` instead of a directory (or directories, for reading)?
> 
> Perhaps to reduce the number of file. We can talk about this on Wednesday. We at some point need to think about the inode limit of the filesystem.

Yes, you're much more an expert here than me but it was on my radar. If the `InstanceDatabase` is handling relocating the save file to a safe location, I was already thinking that we may wish to give that location a nested directory structure based on the node class, e.g. all the runs of `my_module.some_nodes.ThisNode` end up in `.../pwf_read_only/my_module/some_nodes/ThisNode/*`. But basically I was aware the issue may come up and am happy to defer to your expertise on how to solve it.
 
> > * Do we need _both_, or will the `InstanceDatabase` know about where to move the files to?
> 
> Single source of truth would be preferable.

Good, yep, we're on the same page. I think then we may need to rename `InstanceDatabase` as it will have more authority than that implies.

On the topic of extending `InstanceDatabase`'s authority, I was thinking we will probably want to disallow file-caching nodes who don't have a formal version flag. My thought is that we don't want to bother storing MSc Joe's `CustomNode` in the database/safe file location since literally no one will be able to read it. At the same time, file-caching is a nice feature to provide. What do you think about dual support, where you can upload known node results to the centralized database, but are still free to use your local directory as a filesystem-"database" like I do in the example above?

> > * What is the release plan for `pyiron_database` so we can include it as a dependency?
> 
> https://pypi.org/project/pyiron-database/ Currently, I handle postgresql and neo4j as optional dependencies. Seemingly conda is not able to do this.

Can you elaborate? This should be ok, so I think we just need to hammer out the details that are blocking you.

There is no problem specifying optional dependencies in the `pyproject.toml`, although you are correct that users wishing to leverage the optional power will need to manually conda-install those extra dependencies. At test-time we simply use a conda-env file that includes the optional dependencies in order to test the behaviour that relies on them. I guess the only catch is that in the package setup, one needs to make sure that importing the package doesn't automatically trigger attempts to import from the optional dependencies, but this should be manageable by being careful with `__init__.py` files.

`atomistics` is an example with [optional dependencies](https://github.com/pyiron/atomistics/blob/main/pyproject.toml) but a very light [conda recipe](https://github.com/conda-forge/atomistics-feedstock/blob/main/recipe/meta.yaml).

niklassiemer:
We have the package on PyPI already, with optional dependencies. I think it would be quite easy along the lines you mentioned, Liam, to get it to conda-forge as well.
One issue is that the back-ends, as far as I understood Sebastian, are not yet on conda-forge.

XzzX:
> We have the package on PyPI already, with optional dependencies. I think it would be quite easy along the lines you mentioned, Liam, to get it to conda-forge as well. One issue is that the back-ends, as far as I understood Sebastian, are not yet on conda-forge.

Something went wrong in the CI. In principle the dependencies should be there.

`sqlalchemy` - definately
`psycopg2` - definately
`neo4j-python-driver-rust-ext` - made some problems (pip: `neo4j-rust-ext`)

XzzX:
> > > * What is the release plan for `pyiron_database` so we can include it as a dependency?
> > 
> > 
> > https://pypi.org/project/pyiron-database/ Currently, I handle postgresql and neo4j as optional dependencies. Seemingly conda is not able to do this.
> 
> Can you elaborate? This should be ok, so I think we just need to hammer out the details that are blocking you.
> 
> There is no problem specifying optional dependencies in the `pyproject.toml`, although you are correct that users wishing to leverage the optional power will need to manually conda-install those extra dependencies. At test-time we simply use a conda-env file that includes the optional dependencies in order to test the behaviour that relies on them. I guess the only catch is that in the package setup, one needs to make sure that importing the package doesn't automatically trigger attempts to import from the optional dependencies, but this should be manageable by being careful with `__init__.py` files.
> 
> `atomistics` is an example with [optional dependencies](https://github.com/pyiron/atomistics/blob/main/pyproject.toml) but a very light [conda recipe](https://github.com/conda-forge/atomistics-feedstock/blob/main/recipe/meta.yaml).

This is how I already do it.

liamhuber:
@niklassiemer, will you be able to handle releasing `pyiron_database` as a conda package? 

I removed the `bagofholding` dependency here and am just using `pickle` directly -- in the long run we'll still want to convert to `bagofholding` to get version controlling etc., but for the DPG this should be fine. Thus, all that remains is to get access to `pyiron_database.instance_database.get_hash`.

niklassiemer:
I am on it: https://github.com/conda-forge/staged-recipes/pull/29258, the recipe is parsing - we just need to wait for someone from the conda-forge to review and hopefully merge it :)

niklassiemer:
Finally it is on conda-forge 🥳 However we have to get a release with broader dependencies pretty fast I assume (currently fixed dependencies in PyPI and conda-forge, needs of course to be adapted).

liamhuber:
Ahh, @niklassiemer, noooo. `pyiron_database` [depends on `pyiron_workflow`](https://github.com/conda-forge/pyiron_database-feedstock/blob/ee04c53549482991a0dc1ce130f5184d28ad7070/recipe/meta.yaml#L29). So there's no way we can depend on it (in its current form) here.

There's a bit of a silver-lining, in that the only place in the codebase the dependency exists is `pyiron_database.instance_database.node`, and except for `pyiron_workflow.NOT_DATA`, it is purely used for type checks. In principle, `NotData` could get moved over to `pyiron_snippets` (IMO it's a kind of useful alternative to `None` to have around, and I even re-created something equivalent for myself over in `bagofholding`), after which we could perhaps make it optional. After that, though, it is really used concretely in `tests/integration/test_postgresql.py`, so at least for the tests we need a workaround.

@pmrv, you said that for the DPG it's enough that we are saving and loading pre-run workflows, yes? We could hack around this circular problem, but I'd rather go slow and find a more robust and graceful solution if we have time.

pmrv:
For DPG I'm a-ok on the saving front.  In the current state you can save/load fully or partially executed workflows and start off in a new process without problem.  That's all I need.

niklassiemer:
I am optimistic that we find a solution to get these working together. And now we can do some more releases if needed :)

PR #601, opened 2025-02-24
Use instance check instead of equality in as_dataclass_node

pmrv:
type(cls) is only equal to type if no other meta class has been used. This includes classes derived from ABC, so it is a bit inconvenient. Using the instance check however seems to work in these cases.


liamhuber (approved):
Yes, good catch. Now downsides are obvious to me, and the upside is clear. Once the CI passes, I'm happy.

pmrv:
Seems codacy is slow today.  Based on its last [known output](https://app.codacy.com/gh/pyiron/pyiron_workflow/commits/506b78144c89e1371d2e9dbde4315aa769629ae4/issues)  I was gonna ignore it anyway as we discussed on zoom, so I'll go ahead an merge.

PR #605, opened 2025-02-25
Build a readiness report from a dictionary

liamhuber:
And add that dictionary to the raised exception when there is a readiness error.

Closes https://github.com/pyiron/pyironFlow/issues/43


pmrv (approved):
Code looks pretty much like what I had in mind, thanks!  (I went to try it, but could get an editable installation setup on my other laptop, I conclude it is too late over here now.)

PR #607, opened 2025-02-26
Expose the For class in the API

liamhuber:
Analogous to how we expose the Function and Macro classes for node designers to use in explicit subclass creation.

FYI @pmrv


pmrv:
Convenient, thanks!

PR #609, opened 2025-03-04
Add pyiron_snippets/pint to build requires

pmrv:
Otherwise pip install . will not run in a clean env.

@liamhuber As per our ealier mails, but not sure if this is how you intend things to be.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,6 +9,8 @@ requires = [
     "typeguard",
     "setuptools",
     "versioneer[toml]==0.29",
+    "pyiron_snippets",
+    "pint",
 ]
 build-backend = "setuptools.build_meta"
 


liamhuber:
Exactly what I meant; does it resolve your local "." Installation issues?

pmrv:
Jup, I managed to install it this way.

@coverage Seems the action is 'just' timing out, but the actual tests pass.

liamhuber (approved):

liamhuber:
> Seems the action is 'just' timing out, but the actual tests pass.

This has been happening more often lately. IMO Jan is correct that we should move to an alternative provider, I just haven't gotten off my butt and done anything about it in pyiron/actions yet.

PR #613, opened 2025-03-05
Caching `For`

### LINKED ISSUE #610: ForNodes don't seem to cache output and have inverse iteration order
I noticed that in the MWE below, the prints in the for node trigger on every pull, but in the range node only once (as intended).  Also the for loop prints the list from the last item backwards.  Afaict, this is not specified and by piping the input through the node output is not ambiguous, but it's a bit non-intuitive.

```python
@Workflow.wrap.as_function_node('y')
def Double(x: int) -> int:
    print(x)
    return x*2

@Workflow.wrap.as_function_node('range')
def Range(max: int) -> list[int]:
    print('MAX', max)
    return list(range(max))

dl = Double.for_node(iter_on='x', output_as_dataframe=False)

r = Range(4)

dl.inputs.x = r.outputs.range

dl.pull()

dl.pull()
```

liamhuber:
Closes #610 

Basically instead of _always_ rebuilding the child graph, we only rebuild it on cache misses. This is still a bit too aggressive of a rebuild, since in principle we only need to rebuild when the lengths of things we're iterating on change. Nonetheless, it's a simple step in the right direction.

@pmrv does the architecture look ok? Also, I'd like to add a test for this, but all I can think of is iterating over `standard_nodes.Sleep` and making sure it's faster the second time, but I don't really want to make such a slow test -- do you see a more clever route?

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -522,6 +522,7 @@ class Node(
             self.inputs.fetch()
 
         if self.use_cache and self.cache_hit:  # Read and use cache
+            self._on_cache_hit()
             if self.parent is None and emit_ran_signal:
                 self.emit()
             elif self.parent is not None:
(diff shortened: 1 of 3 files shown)


pmrv:
For the tests, what about doing something simple like

```python
values = []
@as_function_node
def SideEffectNode(n: int):
  for i in range(n):
    values.append(i)

s = SideEffectNode(4)
s.pull()
s.pull()
self.assertEqual(len(values), 4)
```
inside a sub test?

pmrv (commented):
Code itself seems clean enough, and should do what we want so :+1:

liamhuber:
I'll add the test and merge tonight then

PR #618, opened 2025-03-12
Force-ignore parent caching on data tree runs

liamhuber:
@Tara-Lakshmipathy I don't actually think we should do this

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -676,7 +676,10 @@ class Node(
                         self.parent.automate_execution = False
 
                     self.parent.starting_nodes = data_tree_starters
+                    parent_caching = self.parent.use_cache
+                    self.parent.use_cache = False
                     self.parent.run()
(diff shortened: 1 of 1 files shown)


liamhuber:
If we're going to go this route, I think we rather need to think about what it means to have caching with a composite -- should they count as being able to use their cache if _any_ of their children fails a cache hit? "No" is a reasonable answer to me, and I think it will also resolve your issue over in https://github.com/pyiron/pyironFlow/issues/104#issuecomment-2716095200, but it comes at some extra computational overhead.

liamhuber:
Having thought about this, I think the existing behaviour on `main` is totally fine. If I `pull` on a child node, and its parent sees that it has a valid cache, I feel the correct behaviour is not to run anything but to exploit the existing cache. So as it stands we waste some time wiring up the (potentially)sub-graph for the `pull` request and then re-wiring it back to its original state, but the external behaviour is just exactly what I expect it to be.

PR #619, opened 2025-03-12
Forbid composite cache hits when a child is not hitting

liamhuber:
@Tara-Lakshmipathy, this, however, I could live with. It's a choice, but I'm very open to it.

### DIFF
--- a/pyiron_workflow/nodes/composite.py
+++ b/pyiron_workflow/nodes/composite.py
@@ -589,3 +589,7 @@ class Composite(LexicalParent[Node], HasCreator, Node, ABC):
         for node in self:
             report = node.report_import_readiness(tabs=tabs + 1, report_so_far=report)
         return report
+
+    @property
+    def cache_hit(self):
+        return super().cache_hit and all(n.cache_hit for n in self.children.values())


liamhuber:
Bear in mind that this would make cache checks recursive -- the cheapest cached result now costs as much as a walk over the _entire_ graph to an arbitrary depth. I see the benefit of the behaviour here, but it is not free.

liamhuber:
@niklassiemer, having complex (unpicklable) IO will always sink serializing the entire workflow containing such nodes, so I don't know the degree to which we'll be able to bury them. Nonetheless, your point is really good that this sort of recursive cache checking plays poorly with hash-based retrieval. If we needed to deep check like this, it wouldn't be enough to see that some macro has a valid hash in the hash-based storage, but we'd then also need to somehow verify the caching state of each child! This basically means you'd need to actually reload (or reinstantiate) the entire macro and check on its internal state. Definitely not what we want.

Thanks for bringing this up; this check is even more costly than I had originally anticipated!

I'm hoping #620 can work, but even #618 is better than the attack here, so I'll close this now.

PR #620, opened 2025-03-12
Don't rely on parent to run data tree

liamhuber:
Another alternative to #618 and #619 for more sensible interactions of caching and pulling.

In principle I like this best, but I need to look at what tests are failing and why, and think harder about why I implemented it the way it currently is to start with.

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -657,30 +657,9 @@ class Node(
                 self.signals.disconnect_run()
                 # Don't let anything upstream trigger _this_ node
 
-                if self.parent is None:
-                    for starter in data_tree_starters:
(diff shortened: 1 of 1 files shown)


liamhuber:
Right, ok, I wind up invoking the `self.parent.run()` here to manage the pull inside the scope of the parent because of how signal emission is set up -- if you have a parent and emit signals, you actually go _via_ the parent rather than emitting the raw signal. I don't like the complexity of this, but handling the signal emission in this way is actually very intentional and lets us have all the things like non-failing branches continuing to run to complete as much graph traversal as possible, and keeping the parent means that recovery files save nicely... so overall I don't want to break this.

PR #621, opened 2025-03-12
Emit signals when running a child directly

liamhuber:
[@Tara-Lakshmipathy pointed out](https://github.com/pyiron/pyironFlow/issues/104#issuecomment-2718649284) that running a particular node doesn't actually push downstream when those nodes are inside the scope of a parent. Indeed, he's right and there was no simple flag or whatever to enable this! 

So here `Node.run` is modified to force signals to emit from nodes who don't have a parent (the existing behaviour) OR who do but whose parent is not itself currently running (and thus managing execution flow) (the new behaviour).

### DIFF
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -523,9 +523,9 @@ class Node(
 
         if self.use_cache and self.cache_hit:  # Read and use cache
             self._on_cache_hit()
-            if self.parent is None and emit_ran_signal:
+            if (self.parent is None or not self.parent.running) and emit_ran_signal:
                 self.emit()
-            elif self.parent is not None:
+            elif self.parent is not None and self.parent.running:
                 self.parent.register_child_starting(self)
                 self.parent.register_child_finished(self)
                 if emit_ran_signal:
@@ -554,7 +554,7 @@ class Node(
         run_finally_kwargs: dict,
         finish_run_kwargs: dict,
     ) -> Any | tuple | Future:
-        if self.parent is not None:
+        if self.parent is not None and self.parent.running:
             self.parent.register_child_starting(self)
         return super()._run(
             executor=executor,
@@ -566,7 +566,7 @@ class Node(
 
     def _run_finally(self, /, emit_ran_signal: bool, raise_run_exceptions: bool):
         super()._run_finally()
-        if self.parent is not None:
+        if self.parent is not None and self.parent.running:
             self.parent.register_child_finished(self)
         if self.checkpoint is not None:
             self.save_checkpoint(self.checkpoint)
--- a/pyiron_workflow/node.py
+++ b/pyiron_workflow/node.py
@@ -770,6 +770,34 @@ class Node(
             **kwargs,
         )
 
+    def push(self, *args, **kwargs):
+        """
+        Exactly like :meth:`run` with all the same flags, _except_ it handles an edge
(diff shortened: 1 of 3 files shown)


liamhuber:
There are a handful of other simple places where one needs to add extra conditions to see if the parent is running (i.e. if the parent is running, we want to go via the `register_child_starting/finished/emitting`, but if there is no parent or it's not running, we simply emit). This is a bit of legwork, but conceptually easy, just adding an extra handful of boolean clauses.

However, as is often the case, the flexible `Workflow` kills it all.

```python
import pyiron_workflow as pwf

wf = pwf.Workflow("push_pull")
wf.n1 = pwf.standard_nodes.UserInput(0)
wf.n2 = pwf.standard_nodes.UserInput(wf.n1)
wf.n3 = pwf.standard_nodes.UserInput(wf.n2)

wf.n1.pull()
wf.n2.run()
print(wf.n3.outputs.user_input.value)
>>> NOT_DATA
```

This is because workflows configure their DAG execution on-the-fly at runtime. So `wf.n2.run()` runs n2 just fine, and even emits its signals, but...they have no receivers! This is a complete non-issue for macros, because they use the same `Composite.set_run_signals_to_dag_execution` to set up their signal flow at _instantiation_ rather than runtime, so it's always available.

The way I got around this is by introducing `Node.push`, which is just a super hacky method for getting around this particular annoying edge case: is my parent a `Workflow`? Is it set to `automate_execution`? Then first `self.parent.set_run_signals_to_dag_execution()` and only then `run`. For children of macros, `run` and `push` are thus identical, while for children of `Workflow`, `run` may fail to propagate downstream under edge cases but `push` will always work (which may include _not_ propagating if the signal flow is manually set to not do that).

So, the above would be fixed like

```python
import pyiron_workflow as pwf

wf = pwf.Workflow("push_pull")
wf.n1 = pwf.standard_nodes.UserInput(0)
wf.n2 = pwf.standard_nodes.UserInput(wf.n1)
wf.n3 = pwf.standard_nodes.UserInput(wf.n2)

wf.n1.pull()
wf.n2.push()
print(wf.n3.outputs.user_input.value)
>>> 0
```

On the one hand, this is obviously a hack. It is ugly and I dislike it. On the other hand, I didn't want to add the boolean check/+flow configuration to each and every `run` call for performance reasons. I would very much appreciate feedback on which of these two others find to be the lesser of two evils -- or even better, if you have a clever idea (preferably including an implementation) for how to avoid the evil altogether.

Tara-Lakshmipathy:
My perspective is a bit biased by the GUI. So, my opinion is that the ugly hack is totally fine because the GUI would only either be using (`pull` and `push`) or (`pull` and `run`) for the top level workflow. As far as I can tell, the main source of performance penalty in the GUI is from passing the JSON object between the view and the model. Compared to that, a performance hit by performing a `push` every time should not be noticable (at least I hope not). So, if you want to keep a low performance-penalty option available for the pure code side (i.e., `run` which does not trigger `set_run_signals_to_dag_execution` and boolean check of the parent), then I think that is totally fine.

Side note, I'm thinking of calling the `pull` button as simply "Run up to" and the `push` button as "Run from". Can't think of anything better right now. Suggestion are welcome and I will create an issue for the naming in the `pyironflow` repo once the features have been introduced.

Tara-Lakshmipathy (approved):
lgtm!

PR #623, opened 2025-03-25
Inline Function class creator

### LINKED ISSUE #622: Is it possible to simplify the mapping of python packages to function nodes
**Summary**

Currently to integrate a new function it is required to write:
```python
import numpy as np
from pyiron_workflow import as_function_node

@as_function_node("sin")
def Sin(x: list | np.ndarray | float | int):
    return np.sin(x)
```
I would like to simplify this syntax to: 
```python
import numpy as np
from pyiron_workflow import as_function_node

Sin = as_function_node(np.sin, "sin")
```
Is there any alternative Syntax available? 

**Detailed Description**

The advantages of the reduced syntax is on the one hand that less lines / characters are required to write new nodes and more importantly for standard functions the signature of the function can be maintained from the source package. The second part applies especially to the case of one of the wrapped Python packages changing the signature of one of its functions. In that case the new solution would automatically reflect the updated signature without the need to manually update the node. In addition to the type hint also the doc string is kept and the source code is accessible to help users to deep dive into a given node. 

**Further Information, Files, and Links**

It seems like this functional was partially available in a previous release:
https://github.com/pyiron/node_library/blob/main/node_library/jnmpi_nodes/atomistic/structure/build.py

```python
from pyiron_atomistics import _StructureFactory
from pyiron_workflow.workflow import Workflow


Bulk = Workflow.wrap.as_function_node("structure")(_StructureFactory().bulk)
Bulk.__name__ = "Bulk"
Bulk.__module__ = __name__
```

liamhuber:
Sometimes we don't have access to the function we want to node-ize to decorate it, and writing out a wrapping function can be annoying, so (under a subset of circumstances) this is a shortcut:

```python
>>> import numpy as np
>>> from pyiron_workflow.nodes.function import to_function_node
>>>
>>> Trapz = to_function_node("Trapz", np.trapz, "trapz")
>>> Trapz.preview_io()
{'inputs': {'y': (None, NOT_DATA),
  'x': (None, None),
  'dx': (None, 1.0),
  'axis': (None, -1)},
 'outputs': {'trapz': None}}
```

- The function must be inspectable
    - e.g. `numpy.arange` fails this requirement
- The function must not use protected or argument names
    - e.g. variadics `*args` and `**kwargs`
- ~The function's annotations must be universally accessible~
    - ~In most cases this means they are restricted to no annotation at all, or a
        type hint from the `builtins` like `int` etc.~

Because of these limitations, I'm leaving it out of the root API for now.

Closes #622

### DIFF
--- a/pyiron_workflow/__init__.py
+++ b/pyiron_workflow/__init__.py
@@ -45,7 +45,12 @@ from pyiron_workflow.logging import logger
 from pyiron_workflow.nodes import standard as standard_nodes
 from pyiron_workflow.nodes.composite import FailedChildError
 from pyiron_workflow.nodes.for_loop import For, for_node, for_node_factory
-from pyiron_workflow.nodes.function import Function, as_function_node, function_node
+from pyiron_workflow.nodes.function import (
+    Function,
+    as_function_node,
+    function_node,
+    to_function_node,
+)
 from pyiron_workflow.nodes.macro import Macro, as_macro_node, macro_node
 from pyiron_workflow.nodes.transform import (
     as_dataclass_node,
--- a/pyiron_workflow/mixin/preview.py
+++ b/pyiron_workflow/mixin/preview.py
@@ -111,6 +111,8 @@ class ScrapesIO(HasIOPreview, ABC):
         thus be left static from the time of class definition onwards.
     """
 
+    _extra_type_hint_scope: ClassVar[dict[str, type] | None] = None
+
     @classmethod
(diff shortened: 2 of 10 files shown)


liamhuber:
#625 gives a way to provide the non-builtin typehints to the in-line converter. It now has a different syntax but otherwise identical constraints as the decorator (single return, no protected argument names/variadics), so I went ahead and added it to the API.

PR #632, opened 2025-05-01
Remove the entire node directory

liamhuber:
In one swoop. This accommodates the CI which is (for reasons yet unclear to me) sticking additional information inside the node directory.


liamhuber:
Docs are still failing. This shouldn't be necessary. Something's gone wrong upstream in the CI. Logging the solution in #633

PR #633, opened 2025-05-01
Just trying to get the CI docs build to trigger

liamhuber:


liamhuber:
Yes, the CI itself is what's failing here, without _any_ changes to the repo relative to two weeks ago when #627 ran fine (modulo coveralls timing out) and five days ago when the dailies ran fine.

I'm logging the solution in #634

Issue #634, opened 2025-05-01
CI failure debugging


liamhuber:
Uploading to coveralls has been spotty for a while, but four days ago the daily test started to fail consistently with:

```
 ======================================================================
ERROR: test_draw (unit.test_node.TestNode.test_draw)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/test_node.py", line 347, in test_draw
    p.unlink()
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/cached-miniforge/my-env/lib/python3.12/pathlib.py", line 1342, in unlink
    os.unlink(self)
IsADirectoryError: [Errno 21] Is a directory: '/home/runner/work/pyiron_workflow/pyiron_workflow/start/cached-miniforge'

----------------------------------------------------------------------
Ran 224 tests in 40.683s
```

Everything is fine on my local machine, this test only produces the three image files and these get cleaned up ok. On the CI it really is putting conda env information _inside_ the node directory. This isn't just an unlucky overlap of paths, in #631 I played around with it and it's consistently and exactly putting it where the node directory is created. Force-removing the whole directory with `shutil.rmtree(...)` does the trick (cf. #632), but really shouldn't be necessary.

What's more, even if I fix that, the docs CI job is still failing:

```
Run mkdir public_html
/home/runner/work/_temp/54c68941-f48b-40bf-89f5-3538d59d9f93.sh: /home/runner/work/pyiron_workflow/pyiron_workflow/cached-miniforge/my-env/bin/sphinx-build: cached-miniforge/my-env/bin/python3.12: bad interpreter: No such file or directory
```

That isn't part of the dailies, but at the latest was working two weeks ago when executorlib was bumped (#627).

So, it really looks like something took a wrong turn in the upstream CI env management exactly between 4 and 5 days ago.

liamhuber:
The dailies are still running fine for other projects using `pyiron/actions`. Thankfully, there is one project under very active development in this period -- `semantikon`. And indeed, looking at the CI logs there, I can see that the push-pull was passing build-docs five days ago, but all the more recent runs fail to build-docs with (modulo pathing) the _exact_ same message I'm getting here:

```
Run mkdir public_html
[...]

PR #636, opened 2025-05-01
Replace all the tests with something super simple

liamhuber:
To try to track down what is going wrong upstream with the placements of conda env files. Trying to make progress on #634 .


liamhuber:
Ok, beautiful, this works fine:

```
Run conda-incubator/setup-miniconda@v3
  with:
    python-version: 3.12
    channels: conda-forge
    conda-remove-defaults: true
    channel-priority: strict
    activate-environment: my-env
    environment-file: ./environment.yml
    auto-activate-base: true
    auto-update-conda: false
    remove-profiles: true
    conda-solver: libmamba
    clean-patched-environment-file: true
    run-post: true
  env:
    PYTHONPATH: /home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit:/home/runner/work/pyiron_workflow/pyiron_workflow/tests/static:/home/runner/work/pyiron_workflow/pyiron_workflow/tests/integration:/home/runner/work/pyiron_workflow/pyiron_workflow/tests/benchmark:/home/runner/work/pyiron_workflow/pyiron_workflow/tests:
    WRITE_ENVIRONMENT_OUTPUT_ENV_FILE: ./environment.yml
```

Now let's turn on miniforge and see what happens

liamhuber:
I had an upstream typo in deactivating the `defaults` channel, but other than that, I can see that the `conda info` are extremely similar, except the path for the location of the env is `` in the failed pyiron/actions usage, and simply `` in my direct usage. 

When I look at the respective invocations of setup-miniconda, the difference is in `activate-environment: my-env` vs `activate-environment: cached-miniforge/my-env`. Let's try to narrow down if this is the cause by modifying the working version to have this path-like env name and try to induce a failure.

liamhuber:
Perfect! Yes, then it looks like the trouble is that I'm providing a path-like env name.

This is interesting, because setup-miniconda [explicitly claims to handle path-like values](https://github.com/conda-incubator/setup-miniconda/blob/142632f154bced74dde05dfc628e82581df045ac/action.yml#L83-L93). Again though, since that code hasn't been touched in four months and the problem is only five days old, maybe it still lies somewhere else.

I also find it further strange, because the [failed mamba example](https://github.com/conda-incubator/setup-miniconda/actions/runs/14697943353/job/41242386219#step:3:858) _also_ started 5 days ago, but there the `activate-environment: anaconda-client-env` isn't path-like at all. Maybe five days ago there was some larger update that is malfunctioning in multiple independent ways?

liamhuber:
Both de-pathing the environment name (`env-path: cached-miniforge_my-env` in the script using pyiron/actions cached-miniforge) and localizing the provided path (`activate-environment: ./cached-miniforge/my-env`) worked fine. 

Interestingly, in the latter case, the conda info is still the big full path:

```
    active environment : /home/runner/work/pyiron_workflow/pyiron_workflow/cached-miniforge/my-env
    active env location : /home/runner/work/pyiron_workflow/pyiron_workflow/cached-miniforge/my-env
```

This is the _exact_ path when we didn't localize and failed, so conda info is sanitizing the path sensibly, but I guess some other step in the process is not. That is so bizarre that I'm going to revert the `./` one more time just to convince myself I can re-break it.

liamhuber:
Ok, this informs me how to modify https://github.com/pyiron/actions/pull/145 to get around this problem. However, I still have no idea about the docs job, and something still changed in a negative way somewhere upstream. So for the moment let's sit and wait.

liamhuber:
Alright, using default parameters works so https://github.com/pyiron/actions/pull/145 is fixing half of https://github.com/pyiron/pyiron_workflow/issues/634 (I still don't know what's going on with building the docs)

PR #638, opened 2025-05-07
Target latest pyiron/actions

liamhuber:


liamhuber:
the build-notebooks and unittests part of the centralized push-pull repo were both failing due to some upstream conda issues (#634) that I hope I patched in https://github.com/pyiron/actions/pull/145. If they pass ok here, then that is final confirmation that the patch worked and also that [the new tag routine](https://github.com/pyiron/actions/issues/142#issuecomment-2859228443) is working ok.

liamhuber:
This ran into the windows-with-an-executor problem (https://github.com/pyiron/pyiron_workflow/issues/48) again for the first time in a long time, but otherwise went fine.

PR #643, opened 2025-05-09
[major] Api split

liamhuber:
In accordance with [25-001-split-api](https://github.com/pyiron/decisions/blob/main/decisions/25-001-split-api.md).

The question then remains, what should be part of which API? I was actually pretty liberal here in leaving things as part of the main user API. I've only put deeper cuts into the dev api file, stuff that would be used for making more complex nodes, catching failures, etc. We still support single-point-of-entry by giving access to the rest of the user API directly in `Workflow`:

```python

def MyNode(x):
    return x + 1

wf = Workflow("my_wf")
wf.n = MyNode(42)
wf.add = Workflow.create.std.Add(wf.n, 0)
```

But honestly, I much prefer the more pythonic numpy-like behaviour of importing the module and being able to directly access the key tools:

```python
import pyiron_workflow as pwf

@pwf.as_function_node
def MyNode(x):
    return x + 1

wf = pwf.Workflow("my_wf")
wf.n = MyNode(42)
wf.add = pwf.std.Add(wf.n, 0)
```

Also note in these examples that I've transitioned both `pwf.standard_nodes` and `pwf.Workflow.create.standard` to `.std`

Feedback on what is nice to have in which API would be appreciated.


ligerzero-ai:
I prefer the second way of doing things, just because there is one extra tab completion on the first.

samwaseda:
I don't have a strong opinion

pyiron_workflow/api.py
```diff
@@ -0,0 +1,19 @@
+"""
+Developer-focused API.
+
+Public items included here are subject to semantic versioning.
+"""
+
+# Node developer entry points
+from pyiron_workflow.channels import NOT_DATA
+from pyiron_workflow.find import (
+    find_nodes as _find_nodes,  # Not formally in API -- don't rely on interface
```
samwaseda, line 10: Since the intention of `api.py` is to make a separate interface to developers, this name change is maybe not necessary?

liamhuber: The change here is to indicate that it's not protected by semantic versioning, so it's still valid. [Both api's are otherwise protected.](https://github.com/pyiron/decisions/blob/1ff6601b45e22a9e97483123a5c2146a1704b9a6/decisions/25-001-split-api.md?plain=1#L18)

Tara-Lakshmipathy:
I also prefer the second way for the reason @ligerzero-ai mentioned. But I'm not sure which stuff should belong to which API. The reason for this is because I actually see 3 levels of interactions with certain overlaps: users, node-developers, module-interface developers (e.g., me developing `pyironflow`). The last level involves access to some stuff that `pyiron_workflow` considers internal at the moment (e.g., `node._cached_inputs`). If these were brought forward, where would they go?

liamhuber:
Interesting, so that's three in favour of the numpy approach over the pyironic single-point-of-Access vs one agnostic. Perhaps it's time to reconsider this point in the zen of pyiron?

The stuff you're talking about, I think, would go in the api. I'm also not sure user/node developer is the right dividing line, as I would still expect any code-based (ie not exclusively gui) user to be writing their own simple nodes. The mental split I was trying to make was "is this user someone who works in a notebook or an ide?" But I agree it is a difficult subjective question what belongs where.

liamhuber:
Perhaps we should amend the decision so that the two sets are not disjoint, but the init content is a pure subset of the api file? This way we could promote/demote stuff to/from the init with a deprecation phase.

Tara-Lakshmipathy:
> Perhaps we should amend the decision so that the two sets are not disjoint, but the init content is a pure subset of the api file? This way we could promote/demote stuff to/from the init with a deprecation phase.

Sounds good. 

>  I would still expect any code-based (ie not exclusively gui) user to be writing their own simple nodes.

Sure, but my question would still remain: should we make a distinction between simple node developers and advanced node developers. But, this is not important.

Anyway,  I'm in favor of doing the split for the developers, with the init being a strict subset of the api file.

liamhuber:
@Tara-Lakshmipathy (and anyone else) I'd love some quick feedback on whether you think I've accurately captured the "most used"/"most friendly" tools in the user `__init__.py` api.

ligerzero-ai:
> @Tara-Lakshmipathy (and anyone else) I'd love some quick feedback on whether you think I've accurately captured the "most used"/"most friendly" tools in the user `__init__.py` api.

LGTM - I mostly just use `as_function_node`

Tara-Lakshmipathy:
LGTM!

PR #653, opened 2025-05-29
[patch] Satisfy updated mypy and configure codecov

liamhuber:
In #652, I discovered that the pypi version of mypy is beyond what is available on conda. I don't want to mess with my local env by pip installing stuff, so I'm going to take a crack at resolving the new complaints without being able to run locally

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,10 @@
+coverage:
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 0.5
+    patch:
+      default:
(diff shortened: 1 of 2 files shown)


liamhuber:
Ok good, this did resolve most of them. The one left -- ` pyiron_workflow/nodes/macro.py:270: error: "Channel[Any]" has no attribute "value_receiver"  [attr-defined]` -- is absolutely correct, I just need to think a bit how to fix it without overly complicating other stuff...

PR #666, opened 2025-06-17
Don't allow injection on "shape"

### LINKED ISSUE #665: Unparented function nodes have leaky output
Consider the following simple setup:

```python
import pyiron_workflow as pwf

@pwf.as_function_node
def Foo(x):
    y = x + 1
    return y

@pwf.as_macro_node
def Bar(self, x):
    self.foo = Foo(x)
    return self.foo

n = Foo()
m = Bar()

wf = pwf.Workflow("some_parent")
wf.n = Foo()

print(
    len(n.outputs.y.connections),
    len(m.outputs.foo.connections),
    len(m.foo.outputs.y.connections),
    len(wf.outputs.n__y.connections),
    len(wf.n.outputs.y.connections),
)
```

As we would expect, this prints `0 0 0 0 0`. _But_, if we run the print again in another cell, it gives `2 0 0 0 0`. And so on, increasing by 2 each call. Each time, the unparented function node is getting new connections to floating nodes of the form `"/injected_GetAttr_m5325131784752424021.obj"`. This is not happening when it's inside the macro, nor inside a workflow, neither is it happening to the macro or workflows themselves. It only happens to the orphan function node.

liamhuber:
Closes #665 

But honestly, at this point I just want to get rid of output injection 🤦‍♂️

### DIFF
--- a/pyiron_workflow/draw.py
+++ b/pyiron_workflow/draw.py
@@ -355,7 +355,7 @@ class Node(WorkflowGraphvizMap):
         )
 
         if depth > 0:
-            from pyiron_workflow.nodes.composite import Composite
+            from pyiron_workflow.nodes.composite import Composite  # noqa: PLC0415
 
             # Janky in-line import to avoid circular imports but only look for children
             # where they exist (since nodes sometimes now actually do something on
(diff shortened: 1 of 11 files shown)


liamhuber:
Something appears to be janky on the github end -- the last commit never ran the tests, and now my latest pushed commit is not showing up

PR #678, opened 2025-06-19
[WIP] Wrap executorlib executors

liamhuber:
To exploit the new caching interface provided in `executorlib-1.5.0` so that nodes can rely on their running state and lexical path to access previously executed results.

Locally with the `SingleNodeExecutor` everything is looking good, but that doesn't natively support terminating the process that submit the job. I'd like to play around with this using the `SlurmClusterExecutor` on the cluster before making further changes.

### DIFF
--- a/pyiron_workflow/executors/wrapped_executorlib.py
+++ b/pyiron_workflow/executors/wrapped_executorlib.py
@@ -1,0 +1,51 @@
+import inspect
+
+from executorlib import BaseExecutor, SingleNodeExecutor
+
+from pyiron_workflow.mixin import lexical, run
+
+
+class CacheOverride(BaseExecutor):
+    def submit(self, fn, /, *args, **kwargs):
+        """
+        We demand that `fn` be the bound-method `on_run` of a `Lexical`+`Runnable`
+        class (a `Node` is, of course, the intended resolution of this demand).
+        """
+        if (
+            inspect.ismethod(fn)
+            and fn.__name__ == "on_run"
+            and isinstance(fn.__self__, lexical.Lexical)  # provides .as_path
+            and isinstance(fn.__self__, run.Runnable)  # provides .on_run
+        ):
+            cache_key_info = {
+                "cache_key": str(fn.__self__.as_path()),
+                "cache_directory": ".",  # Doesn't matter, the path in the key overrides
+            }
+        else:
+            raise TypeError(
+                f"{self.__name__} is only intended to work with the "
+                f"on_run method of pyiron_workflow.Node objects, but got {fn}"
+            )
+
+        if "resource_dict" in kwargs:
+            if "cache_key" in kwargs["resource_dict"]:
+                raise ValueError(
+                    f"pyiron_workflow needs the freedom to specify the cache, so the "
+                    f'requested "cache_directory" '
+                    f"({kwargs["resource_dict"]["cache_key"]}) would get overwritten."
+                )
+            if "cache_directory" in kwargs["resource_dict"]:
+                raise ValueError(
+                    f"pyiron_workflow needs the freedom to specify the cache, so the "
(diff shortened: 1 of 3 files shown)


newfadel (changes requested):
This is a great step forward, leveraging `executorlib-1.5.0`'s caching interface will be very beneficial for performance and result reproducibility!

Regarding your work with `SlurmClusterExecutor`, it's definitely the right approach to test process termination in a real cluster environment. Once you've had a chance to experiment, I'd be particularly interested in understanding:

1.  **Process Management and Cleanup:** How will `SlurmClusterExecutor` manage the lifecycle of the submitted jobs on the cluster? Specifically, what mechanisms will be in place to ensure proper termination and cleanup of resources, especially if the submitting Python process exits unexpectedly or is terminated?
2.  **Caching Strategy with Distributed Execution:** With the caching now active across potentially multiple nodes, have you considered potential challenges around cache invalidation or consistency if underlying data changes or if a previous computation failed? Are there plans to implement strategies to ensure the cached results remain valid and reliable in a distributed setting?
3.  **Error Handling and Robustness:** For production use on a cluster, robust error handling is crucial. How will the wrapped executors handle common cluster-specific issues like job failures, network interruptions, or resource limits?

Looking forward to seeing the progress on this! 😄

liamhuber:
This is currently working to all attempted behaviour when run together with https://github.com/pyiron/executorlib/pull/712

pyiron_workflow/executors/wrapped_executorlib.py
```diff
@@ -0,0 +1,104 @@
+import inspect
+from collections.abc import Callable
+from typing import ClassVar
+
+from executorlib import BaseExecutor, SingleNodeExecutor, SlurmClusterExecutor
+from executorlib.task_scheduler.file.subprocess_spawner import execute_in_subprocess
+from executorlib.task_scheduler.file.task_scheduler import FileTaskScheduler
```
jan-janssen, line 7: @liamhuber I would like to prevent deep links into the executorlib package, as I only promise the `executorlib.X` and `executorlib.api.X` to remain stable. Can you tell me why you need the `LocalFileExecutor`? I currently do not see any functionality which is not covered by the `SingleNodeExecutor` in a more efficient way. I primarily intended the `LocalFileExecutor` for debugging purposes.

liamhuber: I am also including it for debugging purposes, I don't intend to include it in the api here. If I can get the GitHub CI working with Slurm later I'll probably remove it, but so far every time I've encountered a bug/problem with the slurm executor over on the cluster I've been able to reproduce it with the local file executor, solve that locally, and the solution works for slurm too. So it's super useful to me at the moment but never intended for users.

jan-janssen: Yes, that is exactly the use case it is designed for, maybe I should include it in `executorlib.api` as it is helpful for developers to test their code while it is might be confusing for users. In addition it might also make sense to adjust the name, I thought about `TestFileExecutor` or `LocalClusterExecutor`. The first as a reference that it should only be used for testing while the second is more focused on highlighting the similarity to the other cluster executors.

jan-janssen: I opened a corresponding pull request in the executorlib repository https://github.com/pyiron/executorlib/pull/714

liamhuber:
> ImportError: cannot import name 'TestClusterExecutor' from 'executorlib.api' (/home/runner/work/pyiron_workflow/pyiron_workflow/cached-miniforge/my-env/lib/python3.13/site-packages/executorlib/api.py). Did you mean: 'FluxClusterExecutor'?

Waiting on `executorlib-1.5.3`

PR #686, opened 2025-06-26
Bagofholding storage interface

liamhuber:


liamhuber:
```
      ├─ bagofholding =0.0.3 * is installable and it requires
      │  └─ python >=3.12,<3.14 *, which can be installed;
      └─ python =3.10 * is not installable because it conflicts with any installable versions previously reported.
```

Yeah, that is very true.

liamhuber:
> Ok super, I'll touch up the patch coverage tomorrow.

Two of the failures are modified content inside `if typing.TYPE_CHECKING` -- i.e. code that _never_ gets run. We rely on `mypy` for validating the correctness of such code rather than coverage.

I chatted with the AI and the three approaches seem to be:
- Decorate _every line_ with `# pragma: no cover`
  - Abominable.
- Move everything inside a `if typing.TYPE_CHECKING` block into its own module and then ignore this in the `.coveragerc`
  - Interesting. Gets a "free" bump to the coverage rating at the cost of a little complexity
- Just live with it
  - This one. Today, definitely, this one.

PR #711, opened 2025-07-18
Prevent parent/detatched path conflict

liamhuber:
For macros with executors. Closes https://github.com/pyiron/pyiron_workflow/issues/710


liamhuber:
Codecov thinks the patch line is not covered because it doesn't see the process pool executed business. The new test covers exactly this case, and indeed the test fails when the new line is missing.

Issue #722, opened 2025-07-21
Wrapped executor naming?


liamhuber:
@jan-janssen, I forgot to check during the meeting today -- what do you think of the wrapped executor name [`pyiron_workflow.NodeSlurmExecutor`](https://github.com/pyiron/pyiron_workflow/blob/3fa161fe692c3a1547f715e20216402b2f4bcb94/pyiron_workflow/executors/wrapped_executorlib.py#L76)?

I do want to keep a bespoke executor so that all the serialization location and cleanup can be managed, and I internally add silent checks to verify and indicate that this for usage with `pyiron_workflow` nodes, so I'm content with the behaviour. But I want to check with you (a) whether the use of the term `Node` in a new context is too confusing, and (b) whether you want any more explicit attribution to `executorlib` right in the name?

I think it's nice for usability if `Slurm` and `Executor` appear in the class name, but am otherwise completely flexible.

jan-janssen:
The challenging part are the two executors for SLURM, executorlib has the `SlurmClusterExecutor` https://executorlib.readthedocs.io/en/latest/2-hpc-cluster.html#slurm and the `SlurmJobExecutor` https://executorlib.readthedocs.io/en/latest/3-hpc-job.html#slurm . How do you differentiate those two? Are they both named `NodeSlurmExecutor`?

jan-janssen:
I continue to think about it and I wonder, would it be possible to dynamically generate the executor classes in pyiron_workflow? The advantages I see:
* Users could update their workflow by switching from one executor to the next, e.g. from ProcessPoolExecutor to SlurmClusterExecutor.
* If another framework exists which also uses the executorlib pattern users could combine it with pyiron_workflow, here are a couple of similar frameworks I am aware off https://github.com/pyiron/executorlib/issues/630 
* It would be possible to define executorlib as optional dependency for pyiron_workflow, so users who do not need the HPC integration, e.g. experimentalists , would benefit from a smaller installation foot print.

jan-janssen:
I definitely agree with complain one and we had the same discussion in executorlib https://github.com/pyiron/executorlib/issues/703 . The integration of an Executor is not an easy task, so it is highly likely that the internal structure is complex as a result.  Maybe one idea would be instead of:
```python
import executorlib
import pyiron_workflow as pwf

n = pwf.std.Sleep(60)
[...]

PR #726, opened 2025-07-22
Dynamically wrap executorlib classes

liamhuber:
To dynamically `create executorlib` executors that use `CacheOverride` to be specific for `pyiron_workflow`. The hint is probably too generic, as we only expect this to even possibly work for base classes that leverage the file caching functionality.

A demo for #722


liamhuber:
I don't mind just having the static set of executors.

PR #733, opened 2025-07-23
Explore ontology tools with pyiron_workflow

liamhuber:
A notebook-level exploration of how to integrate ontology tools with `pyiron_workflow`. 

Most things are working out-of-the-box, and it looks like it will be simple and pretty robust to attack ontological checking of new connections in the following way:
- Validate types first
- Get the graph-root of the node in question
- Parse the entire workflow graph with `pyiron_ontology.parser.parse_workflow`
  - This should probably be relocated here
  - This doesn't seem to be playing well with all node types, but function nodes, macros, and workflows are all working well so we can at least get started
- Pass the parsed dictionary to `semantikon.ontology.validate_values`
- If there are missing triples or incompatible connections, revert the connection and report them in a raised error

Also on the todo-list is to generate lists of candidate channels to connect to or nodes to add given an ontologically-hinted channel and a corpus of existing channels or nodes (respectively), but let's start with just validating connections.


samwaseda:
> @samwaseda, I made this PR because I was hoping to get some feedback from you. Even if it doesn't execute, at least the pre-executed notebook is readable [here](https://github.com/pyiron/pyiron_workflow/blob/a946ad32c4eb294e5f10fc085f323234fc0bdd56/notebooks/explore_ontology.ipynb).
> 
> First, thank you for the `pyiron_ontology.parser.parse_workflow` work! You can see in the notebook that I'm able to reproduce the `semantikon` [knowledge graph notebook](https://github.com/pyiron/semantikon/blob/main/notebooks/knowledge_graph.ipynb) basically out-of-the-box on `pyiron_workflow` graphs with this! 🚀
> 

🎉

> My main question is whether there is a way to pre-populate the graph with this sort of information, or to include it already in the `Rice` definition so that we can parse and validate the `pyiron_workflow` graph in a single shot?

Maybe I involve @Tara-Lakshmipathy and @srmnitc here, but in principle the graphs are usually pre-populated with this kind of definition, and that's what ontologists should provide us. I guess for this example you can leave the line as it is.

liamhuber:
> I guess for this example you can leave the line as it is.

For the example, yes, but not for the implementation. For the implementation we need something like `ontology.validate_values(parser.parse_workflow(self.owner.graph_root))` inside the channel connection validation, where `self` here is the channel doing the connecting and its `graph_root` is the top-most node available. I.e. we are only getting the graph at runtime.

Now, we could allow users to add ontological restrictions to (arbitrary?) nodes and collect these (by walking a tree?) then break the above into multiple lines

```python
graph = parser.parse_workflow(self.owner.graph_root))
for extra_info in self.owner.recursively_grab_extra_onto_information_or_something():
    graph.add(extra_info)
ontology.validate_values(graph)
```

But I'm wondering if this is a reasonable way to do it?

I also have some concerns about where and how we store this information, since when we run nodes on remote processes we de-parent them first (so that we don't ship the _whole_ graph off to the other process). However, I don't think it's an existential threat as we could always collect and store this information in the state before de-parenting, as long as it's going to be relatively lightweight.

samwaseda:
No, the graph should already be filled with the user-specified ontology, so you wouldn't start writing `graph = parser.parse_workflow(wf)` but instead you'd for example write:

```python
from rdflib import Graph

g = Graph()
foaf_url = "http://xmlns.com/foaf/0.1/"
g.parse(foaf_url)
for triple in parser.parse_workflow(wf):
    g.add(triple)
```

If you prefer, we can also make an example with [CMSO](https://github.com/OCDO/cmso).

> I also have some concerns about where and how we store this information, since when we run nodes on remote processes we de-parent them first

But there is nothing we check during the run-time, so by the time we run them, we can safely de-parent them, right?

liamhuber:
Ah yes, I'm being silly. Indeed we would expect the graph construction to be specified all on the parent process so de-parenting should be ok. There might be some sort of tricky edge cars with dynamic nodes, but even there I would expect the recipe validation to already be finished so new edges should still be compliant.

liamhuber:
And specifying it before then adding the graph parse sounds great. Actually, the presence of that stored graph somewhere in the tree is probably a great flag for whether or not to attempt ontological validation (that at the participants in the edge holding ontological hints)

liamhuber:
A CMSO demo would be rad, but I think for the purposes of building the infrastructure the toy examples are probably already enough to get rolling well.

I'll leave this open for a bit in case the others want to share their perspective, but my immediate questions have been answered. Thanks, @samwaseda!

liamhuber:
It took some minor modification, but it looks like there's not problem including the more complex node types. At the recipe stage none of them have their own subgraph, so I just treat them all as atomic nodes and it works fine. Still, only the `Function` nodes can wrap a declaration with its own `@u` decorator, but most of the others wouldn't come across that anyhow -- e.g. it makes no sense to ontologically declare what a dataclass node is doing, as it is literally just reformulating the data.

The thing that I'm stuck on now is getting macros to validate the IO passing (via `value_receiver`). I can live with it only happening at instantiation time, but right now I can't force a failure with bad ontological hints. This is demonstrated [in the notebook](https://github.com/pyiron/pyiron_workflow/blob/a96cbcf4cb6985a4127f0a697a238bdef479252b/notebooks/explore_ontology.ipynb)

liamhuber:
With the recent merged solutions and the workaround in https://github.com/pyiron/semantikon/pull/247, most of the examples I have here are working ok. What remains is the for-loop (and probably while-loop). I suspect there the issue is the order of creation and labelling.

liamhuber:
```
  error    libmamba Could not solve for environment specs
      The following package could not be installed
      └─ semantikon =0.0.22 * does not exist (perhaps a typo or a missing channel).
```

It's live [on the web](https://anaconda.org/conda-forge/semantikon) but must not have propagated to these servers yet. I'll try again later.

Otherwise this is down to unit tests and the better exception messages -- we handle everything I can get `semantikon` to handle.

liamhuber:
Ahaaa, it is not yet all alright. `semantikon` parses and validates complete graphs at the end of their life cycles. `pyiron_workflow` lets users add nodes and then wire them up later.

So a simple example like this:

```python
import rdflib
from semantikon.metadata import u

import pyiron_workflow as pwf

EX = rdflib.Namespace("http://example.org/")

@pwf.as_function_node
def Up(x: u(str, uri=EX.TriggerOnto)) -> u(str, uri=EX.TriggerOnto):
    return x

@pwf.as_function_node
def Middle(y: u(str, uri=EX.TriggerOnto)) -> u(str, uri=EX.TriggerOnto, triples=(EX.hasThing, EX.thing)):
    return y

@pwf.as_function_node
def Down(
    z: u(
        str,
        uri=EX.TriggerOnto,
        restrictions=(
            (rdflib.OWL.onProperty, EX.hasThing),
            (rdflib.OWL.someValuesFrom, EX.thing),
        )
    ),
):
    return z
```

Works fine when we do things in the right order:

```python
wf = pwf.Workflow("connect_first")
wf.up = Up("this is fine")
wf.middle = Middle(wf.up)
wf.down = Down(wf.middle)
wf()
>>> {'down__z': 'this is fine'}
```

But if our node with restrictions dares to even exist before we try forming an unrelated connection, the current setup complains because the validation of the entire graph is failing

```python
wf = pwf.Workflow("connect_later")
wf.up = Up("this is is not")
wf.middle = Middle()
wf.down = Down()
wf.middle.inputs.y = wf.up
>>> ChannelConnectionError: The upstream channel /connect_first/up.x cannot connect to the downstream channel /connect_first/middle.y because the upstream type hint (typing.Annotated[str, ('uri', rdflib.term.URIRef('http://example.org/TriggerOnto'))]) and downstream type hint (typing.Annotated[str, ('uri', rdflib.term.URIRef('http://example.org/TriggerOnto'))]) produce a non-empty ontological validation report:
... {'missing_triples': [(rdflib.term.URIRef('connect_first.inputs.z'), rdflib.term.URIRef('http://example.org/hasThing'), rdflib.term.URIRef('http://example.org/thing')), (rdflib.term.URIRef('connect_first.down.inputs.z'), rdflib.term.URIRef('http://example.org/hasThing'), rdflib.term.URIRef('http://example.org/thing'))], 'incompatible_connections': [], 'distinct_units': {}}
```

What we really need to check in this case is that the new connection is not involved in any of the problems.

liamhuber:
All the edge cases I've been able to think of are finally handled the way I want (given https://github.com/pyiron/semantikon/issues/262), and the suggestion "tests" are printing what I want to see. Fingers crossed, it's just a matter of turning the print statements into real tests and revising the pedagogical notebook one last time.

Issue #735, opened 2025-07-29
Handling of directories


samwaseda:
One thing I really like about `pyiron_base` is that I don't need to worry much about the directory structure, i.e. I simply write the files in `self.working_directory`, and afterwards when the job is erased there is no orphaned job directory. So I would like to suggest a couple of things here purely from the user perspective first, without distinguishing what should be offered by us and what the user should provide.

Let's assume I have a software tool that requires input files and exports output files. That means there's an input writer, and an output parser. Since I'm interested in the input parameters that I define, and the output parameters that I choose, I'm not really interested in the actual files, but just in case I would like to be able to look into them. That means I would need the following features:

1. A unique job folder should be generated for each job.
2. The job folder should in principle follow the same fate as the job itself, meaning it shouldn't be orphaned when the job is erased
3. It should lie inside the workflow folder, which is currently not automatically generated until `save` is called (and also `mkdir` is hard coded if I remember correctly)
4. I would like the WfMS to compress everything, since I usually don't need the files (`compress` & `decompress` from `pyiron_base`)

I'm getting the feeling that the right moment to create the workflow folder is maybe when `as_path` is called, and if it remains empty we can potentially erase it again after the run.

I guess I'm not totally done with this part, but I'm gonna come back to this and add more comments if something comes to my mind.

liamhuber:
I'm fully on board with looking for ways to automate away more file system stuff.

> So I would like to suggest a couple of things here purely from the user perspective first, without distinguishing what should be offered by us and what the user should provide.

On this note, there's also a middle ground where we do stuff in the standard corpus of nodes, but not in the engine itself. There's no guarantee we'll be able to solve our problems this way, but I like it because it provides a very natural way to extend power without adding complexity to the core engine.

> It should lie inside the workflow folder

This is the tricky thing. Ultimately, we are pushing towards a function-based approach, and the writing and reading should be done by raw functions. These must necessarily take their working directories as inputs. While the operation of the function doesn't (indeed mustn't) know that it belongs to a node and a graph, we as users sure know it does and would love to exploit that.

It might seem at first that macros help us out of this, since they know about the fact that they are nodes -- indeed, the current implementation currently takes the `self` argument of the macro during the definition. Something to remember though, is that traditional macro definitions are recipe declarations and get run at instantiation (so that we have access to the graph object as early as possible) and do not get re-run at runtime.

[...]

PR #736, opened 2025-07-29
Add `mkdir` to `as_path`

samwaseda:
First item in the list of #735

### DIFF
--- a/pyiron_workflow/mixin/lexical.py
+++ b/pyiron_workflow/mixin/lexical.py
@@ -143,14 +143,17 @@ class Lexical(UsesState, HasLabel, Generic[ParentType], ABC):
         else:
             return self
 
-    def as_path(self, root: Path | str | None = None) -> Path:
+    def as_path(self, root: Path | str | None = None, mkdir: bool = False) -> Path:
         """
         The lexical path as a :class:`pathlib.Path`, with a filesystem :param:`root`
         (default is the current working directory).
         """
-        return (Path.cwd() if root is None else Path(root)).joinpath(
+        path = (Path.cwd() if root is None else Path(root)).joinpath(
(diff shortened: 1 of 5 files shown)


samwaseda:
I thought this was a pretty neat implementation, but then with the following code I encountered a problem:

```python

import uuid
from pathlib import Path

def generate_working_directory(path):
    working_directory = Path(path) / Path(uuid.uuid4().hex)
    working_directory.mkdir(exist_ok=True, parents=True)
    return working_directory

def write_hello(working_directory, file_name="my_file.txt"):
    full_path = Path(working_directory) / Path(file_name)
    with open(full_path, "w") as f:
        f.write("hello")
    return full_path

@Workflow.wrap.as_macro_node
def example_macro(macro, file_name="my_file.txt"):
    path = macro.as_path(mkdir=True)
    macro.working_directory = generate_working_directory(path)
    macro.message = write_hello(working_directory=macro.working_directory, file_name=file_name)
    return macro.message

wf = Workflow("wf")
wf.macro = example_macro()
```

This created a directory called `example_macro`, and not `wf/example_macro`. Is there a way to map the full hierarchy? @liamhuber

liamhuber (approved):
@samwaseda there is absolutely nothing wrong with this, but I don't yet see how we will use it so my brain is waving a  YAGNI yellow flags are waving.

samwaseda:
> I don't yet see how we will use it so my brain is waving a YAGNI yellow flags are waving.

ok let's see how people are gonna react, but I'm relatively sure that people will like to have a straightforward way to get a folder where they can freely store whatever they want. At the same time it was a bit confusing for me that `as_path` gives me a path, but the path actually doesn't exist, so even if the default behavior is going to be the same, I think it makes sense that there's a way to make the path exist. Anyway since you approved it I merge it 😎

PR #737, opened 2025-07-29
Force delete with `delete_even_if_not_empty`

samwaseda:
Another feature mentioned in #735

I made the argument particularly long and keyword-only (at least from the user interface). I guess the functionality is covered in the tests.


samwaseda:
Did slurm fail because of me?

liamhuber (approved):
LGTM.

From the codecov report it looks as though the case where things are already non-existent is not tested. I'd be happier if this edge were covered, but it's otherwise good.

I find the slurm setup action sometimes times out before we ever get to running our tests, so this can just be re-run.

Issue #32, opened 2023-10-16
Make node replacement even easier


liamhuber:
@JNmpi and I had a super short chat after the pyiron meeting tomorrow, and he was saying he'd still like to see syntax like:

```python
# wf.min_phase1.calc = Macro.create.atomistics.CalcStatic  # Currently possible
wf.min_phase1.calc.replace_with.static()  # TODO, or similar
```

After introducing replacement and the `Node.replace_with(self, other: Node)` method in #26, this should actually be super easy to implement! A sketch looks like this: replace the `replace_with` method on `Node` with a property returning an instance of the following class:

```
class Replacer:
    def __init__(self, node: Node):
        self.node = node

    def __call__(self, other: Node):
        # This is just the current replacement code:
        if self.parent is not None:
            self.node.parent.replace(self, other)
        else:
            warnings.warn(f"Could not replace {self.node.label}, as it has no parent.")
```

Then a savvy developer can "power-up" their node with replacements they know to be compatible by overriding the `replace_with` property with an instance of a child of that class like:

```
class CalculatorReplacer(Replacer):
    @property
    def minimize(self):
        self(Workflow.create.atomistics.CalcMinimize)

    @property
    def static(self):
        self(Workflow.create.atomistics.CalcStatic)

    # Etc.
```

If the replacement is invalid this will just fail the way replacement normally fails -- e.g. if you have a minimize  node with `pressure` connected to something, replacing with `static` should fail. I might need to double-check the tests and failure cases here to be sure this is working as intended, but the idea is perfectly straightforward.

As described above, this only works for nodes which are defined by actually making subclasses. We could probably stretch it to work using the decorators by extending the signature of `Node` to include `replacer: Optional[type[Replacer]]` and accept that as a decorator argument. Still requires the dev to be sophisticated enough to define a child-class of `Replacer`, but at least the decorators could still be used something like:

```python
@Workflow.wrap_as.function_node(replacer=CalculatorReplacer)
def calc_static(job):
    # Blah blah
```

A bit of leg-work to get the examples and tests all set up, but I'm pretty happy with this direction.

PR #743, opened 2025-08-13
No node replacement

liamhuber:
This PR removes the `wf/macro.replace_child(...)` and `node.replace_with(...)` methods entirely. 

For macros, I think these are really the devil. Macros should be clearly defined static recipes, and not get tampered with. In the future I'd like to lock them down.

For workflows, it's potentially a little annoying to lose this functionality. However, at the development stage, I think the vast majority of workflows will be cheap enough to simply re-execute from scratch. In the worst case, you can still remove the child and add the new one under the same name; re-connection can be done manually, or by first disconnecting the child to be removed and leveraging the connections returned there.

Overall, I don't get the feeling this feature is seeing much use and I will be surprised if it is missed. If it is, the `test_macro.TestMacro.test_storage_for_modified_macros` gives a hint of a way to re-introduce it, although that lacks the recreation of the value receiver links.


liamhuber:
Closes #32 

By way of getting rid of the concept

PR #744, opened 2025-08-13
Only intra-graph connections

### LINKED ISSUE #587: No scope-breaking connections, please
IMO, data should only flow between nodes in the same graph (i.e. the same parent scope). Practically, we behave this way, but there is no enforcement to _stop_ bad behaviour. I.e. this works and I don't think it should:

```python
import pyiron_workflow as pwf

@pwf.as_macro_node
def MyMacro(self, a: int, b: int) -> int:
    self.c = a + b
    return self.c

wf = pwf.Workflow("demo")
wf.one = MyMacro()
wf.two = MyMacro()

wf.one.c.outputs.add.connect(wf.two.c.inputs.obj)
```

Because `wf.one == wf.one.c.parent != wf.two.c.parent == wf.two`.

liamhuber:
Enforce the exclusive connection of nodes with the same (or `None`) parent.

This was always the intent. In order facilitate passing connections (but not a `parent=`) at instantiation time, when connections are made from a parented node to an orphan node, the orphan gets its counterpart's parent. This results in one gross side effect: making connections from an un-parented node to a node _inside_ a macro _after_ the macro has finished instantiating adds the un-parented node as a child of the macro! This has no functional impact since the execution graph of the macro is not updated post-instantiation, so that node just sits there taking up memory, but it is still sub-optimal.

Closes #587

### DIFF
--- a/notebooks/deepdive.ipynb
+++ b/notebooks/deepdive.ipynb
@@ -17,15 +17,15 @@
    "id": "771da684-348b-47e2-ba7c-3c2244e46ba3",
    "metadata": {
     "ExecuteTime": {
-     "end_time": "2025-06-18T17:46:40.938676Z",
-     "start_time": "2025-06-18T17:46:40.935402Z"
+     "end_time": "2025-08-13T22:35:55.690634Z",
+     "start_time": "2025-08-13T22:35:55.687720Z"
     }
    },
    "source": [
     "import pyiron_workflow as pwf"
    ],
    "outputs": [],
-   "execution_count": 45
+   "execution_count": 81
   },
   {
    "cell_type": "markdown",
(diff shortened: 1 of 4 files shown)


liamhuber:
The missing coverage is actually relevant; before merging I should extend the test to include such a callback-triggering failure case.

liamhuber:
Also let's replace the callback class with `with contextlib.ExitStack()`

liamhuber:
> Also let's replace the callback class with `with contextlib.ExitStack()`

No, `ExitStack` triggers all its callback _whenever_ we exit the stack context; I need something that only triggers callbacks when we hit an exception inside the context.

I could still use a context manager, but for now I only have a single exception anyway. It might be interesting to replace all such cleanup tasks with an exception-cleanup-manager, but that can be a separate PR.

PR #754, opened 2025-09-12
Update for changes in `semantikon`

liamhuber:
In https://github.com/pyiron/semantikon/pull/266, inheritance of `uri=` is removed when `derived_from=`, and the bug of inheriting `restrictions=` on connections is resolved. Here I make the small adjustments necessary to stay up-to-date with that.

Pending the next `semantikon` release.


liamhuber:
Done in #840

PR #762, opened 2025-10-03
[nit] Keep binder dir clean

liamhuber:
in perpetuity, but using a storage list instead of a deletion list


liamhuber:
I launched binder for the branch, it works fine 🚀

PR #853, opened 2026-05-29
Flowrep-based WfMS

liamhuber:
This PR rebases `pyiron_workflow` onto `flowrep` to follow the paradigm `flowrep.RecipeNode` -> WfMS -> `flowrep.NodeData`. In this way, the WfMS has two key responsibilities:

1) Converting prospective `flowrep` "class-view" recipes into retrospective `flowrep` "instance-view" data. Including, but not limited to:
  - Resource allocation (i.e. executors)
  - Logging, state management, and crash recovery
  - Tracking execution-related metadata that lives outside the scope of the `flowrep.NodeData` structure (e.g. run duration)
2) Facilitating building new `flowrep` recipes
  - Critically, this means the WfMS view must accept _transiently invalid_ recipe states
  - With convenient code syntax (e.g. adding nodes by attribute assignment)
  - Exposing easy handles for GUI wrappers
  - Undo/redo stacks for graph state

While these feautre sets overlap strongly with the existing `pyiron_workflow` functionality, there is a really fundamental difference that the `pyiron_workflow` nodes are no longer the ground truth. On the prospective side, we revert to the `flowrep` recipes being the ground truth. Most of the time, that just means digesting such recipes into WfMS nodes, but in the case of the mutable "transiently invalid" workflow, it means that we dynamically produce such recipes -- either for saving and sharing in their `flowrep` format, or for guiding execution. On the retrospective side, the `pyiron_workflow` nodes are no longer holding mutable data state -- each "run" takes the recipe as-is and produces a `flowrep.NodeData` associated with that recipe. I.e., each run produces a static recipe/data pair artefact (plus other metadata).

### Plan

In light of this fundamental difference, I opted to build the `flowrep`-based WfMS from the ground-up in parallel with the existing implementation. This lives in the `_wfms` submodule (and under the same name in the integration and unit test folders). It is unavoidable that invoking runs, what comes out of those runs, and the (de)serializiation tools will differ, but I'm extremely optimistic that the vast majority of the current API can be maintained. I.e., even though `flowrep` "node decorators" leave the plain function in place and simply add a recipe attribute to it, here we can continue to have `@as_function_node` and `@as_macro_node` decorators which still transform the decorated function into a _node factory_. But now the node produced will be the `flowrep`-based node and not the existing node class instance. In this way, while users will unavoidably need to adjust how they run their workflows, all the workflows they have written down in `.py` files using the current API constructors will remain valid. However, getting this backwards compatibility to work is going to require a bit of hijinks that I don't love, and for long-term maintainability we'll want to drop that compatibility and roll back those hijinks. In the long run, we anyhow want users to be writing their workflows (in `.py` or `.json`) directly as `flowrep` recipes and then digesting those (anyhow, this is already more friendly for flow controls, which can be parsed directly from a python function by `flowrep`!)

Thus, I plan three stages:
1. Introduce the `flowrep`-based WfMS privately alongside the existing implementation (this PR)
2. Swap positions and expose the `flowrep`-based version in the API, and move the existing version to a `legacy` submodule
3. Drop the `legacy` submodule, drop the backwards-compatible API exposures, and roll back any particularly nasty hijinks that were needed for the compatibility

A release at each stage means users would have two versions with co-existence of old and new formats to help smooth migration.

### Features

Gained:
- Full `flowrep` compatibility
  - Consumes recipes
  - Graph state can be exported to recipe
  - Running produces `flowrep` data
- `Run` output container provides a place to store additional metadata
- Macro/Workflow equivalence and bidirectional conversions
  - Facilitates un/grouping
  - Round-trip is necessarily lossy on the underlying python function reference (if any) for macro->workflow->macro
- _Really_ easy graph mutation handles for GUIs (`add/remove/rename_node/edge`, `un/group`, etc.)
- Runtime hook system
  - Status updates and a status system (pending/running/finished/failed)
  - On-exception hooks, e.g. for dumping run state
    - Even when workflow execution is split across processes (e.g. with `ProcessPoolExecutor`)

Lost:
- Caching
  - The `pyiron_workflow` no longer holds data state, so it doesn't even make much sense
  - Only meaningfully applies to atomic nodes, which in the `flowrep` world stay as plain functions (with a recipe attribute attached), so I hope to offload caching to a third-party and already have a tool in mind
- Invoking a run without providing input data
  - Again, the `pyiron_workflow` nodes themselves no longer store data state, and `flowrep` doesn't allow recipes with inputs that have neither an edge nor a python reference with a default for that field -- so you always need to provide terminal input explicitly
- Building and running a graph of peer-nodes outside the scope of a parent graph object

TODO:
- [x] Ontological validation
  - ~`semantikon` already accepts `flowrep` objects, so I just need to test and debug the interface -- no conceptual concerns~ ~Debugging in https://github.com/pyiron/semantikon/pull/455 https://github.com/pyiron/semantikon/pull/457, just need to write up the tests here~ Done.
- [x] API parity
  - Running and what those runs output will look different, and storage routines have changed, but all (or at least the vast majority of) the workflows written with `@as_function_node`, `@as_macro_node`, and other existing API elements should continue to work but return new `flowrep`-based node objects
    - ~The only part I'm currently concern about is how to capture default values in `@as_macro_node`-decorated functions~ this was resolved via https://github.com/pyiron/flowrep/pull/252
    - EDIT: I still want to catch basic API compatibility in `@as_function_node` and `@as_macro_node`, but let's leave other elements to a future PR. 
- [x] Documentation: a user notebook for the new framework
- Pull mode
  - Involves walking the graph dependency tree and building a new dynamic `Workflow` out of the parts touched
  - EDIT: In-scope for the next release, but probably do it in a separate PR after this
- `executorlib` interface with persistence across session shutdowns
  - We don't have dynamic classes anymore, add the end of the day we're just evaluating plain functions/methods, so combined with independent developments in `executorlib`, I'm optimistic this interface will get both simpler and more robust
  - EDIT: In-scope for the next release, but probably do it in a separate PR after this
 - EDIT: extend the `Run` fields to collaborate with metadata used by `courier`
   - Needs to be there before a 1.0 release, but otherwise timing is super flexible; probably not completely finished in this PR


liamhuber:
`operators.methodcaller/itemgetter/attrgetter` don't expose signatures for python <3.12. I don't want to have the stdlib vary based on intepreter version, so let's not nest the declarations inside a version check; I'm not ready to drop 3.11; that leaves just removing these three from the stdlib.

liamhuber:
`[10](https://github.com/pyiron/pyiron_workflow/actions/runs/27787654161/job/82228427030?pr=853#step:4:1332)
pip._vendor.pyproject_hooks._impl.BackendUnavailable: Cannot import 'hatchling.build'`.

Looking into whether this could possibly be me, or if it's just part of GitHub's degraded performance of late

PR #873, opened 2026-06-23
Dataclass decorator

### LINKED ISSUE #864: `flowrep`-based dataclass QoL
The treatment of dataclasses in `flowrep` is pretty primitive compared to the legacy `pyiron_workflow`. In particular, there's no dataclass->node decorator. Before releasing the new `flowrep`-based WfMS, I want some better QoL stuff here. It could (should?) be mostly right in `flowrep` itself, but it can always start here and migrate relevant portions later.

From my notes yesterday morning>

> dataclasses should be pretty easy.
We can have two functions mapping dataclasses to output ports and vice-versa.
The dataclass decorator can apply the `dataclasses.dataclass` decorator, and then use `functools.partial` to make two new functions (on for each direction) that binds the dataclass itself in, and then we modify the attached signature object according to dataclass fields (in order to get the right IO behaviour).
These functions are then bound to the dataclass object (we'll need to validate that nothing in its dir conflicts first).
Flowrep then parses _those_ functions and adds a recipe to them.
>
> So you'd do like
>
> ```python
> import flowrep as fr
> 
> @fr.dataclass
> class Foo:
>     bar: int
>     baz: str
>
>     def razzle_dazzle(self):
>         return self.bar * self.baz
>
> Foo.dataclass2outputs.flowrep_recipe
> ```
>
> At the `pyiron_workflow` level, we'd just add the `.pwf` constructor to the two attached functions, like usual.

liamhuber:
`@pyiron_workflow._wfms.api.dataclass` operates as a normal `@dataclasses.dataclass` decorator _and_ extends the dataclass with two new fields `.pwf_inputs2dc` and `.pwf._dc2outputs` analogous to the `.pwf` tools added to `@atomic` and `@workflow` decorated functions. These offer the same `.node` and `.run` methods, for transforming dataclass instances to/from IO ports.

Closes #864

### DIFF
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 - bidict =0.23.1
 - cloudpickle =3.1.2
 - executorlib =1.9.3
-- flowrep =0.5.1
+- flowrep =0.5.2
 - graphviz =12.2.1
 - hatchling =1.29.0
 - hatch-vcs =0.5.0
--- a/.binder/environment.yml
+++ b/.binder/environment.yml
@@ -16,7 +16,7 @@ dependencies:
 - python >=3.11,<3.14
 - python-graphviz =0.21
 - rdflib =7.6.0
-- semantikon =1.3.1
+- semantikon =1.3.2
 - toposort =1.10
 - typeguard =4.5.2
 - python-workflow-definition =0.1.5
--- a/.ci_support/environment.yml
+++ b/.ci_support/environment.yml
@@ -5,7 +5,7 @@ dependencies:
 - bidict =0.23.1
 - cloudpickle =3.1.2
 - executorlib =1.9.3
-- flowrep =0.5.1
+- flowrep =0.5.2
 - graphviz =12.2.1
 - hatchling =1.29.0
 - hatch-vcs =0.5.0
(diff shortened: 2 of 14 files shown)


liamhuber:
```
[855](https://github.com/pyiron/pyiron_workflow/actions/runs/28003701794/job/82881142138?pr=873#step:4:1864)
FAIL: test_string_annotation_resolves (unit._wfms.test_decorators.TestDataclass2Outputs.test_string_annotation_resolves)
  File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/unit/_wfms/test_decorators.py", line 118, in test_string_annotation_resolves
    self.assertEqual(node.outputs["a"].type_hint, int)
AssertionError: "'int'" != <class 'int'>
```

This ran fine locally, but failed when I checked out `flowrep-0.5.1` locally too. I think the fix is in https://github.com/pyiron/flowrep/pull/261. It is fine to just wait until [`flowrep-0.5.2`](https://github.com/pyiron/flowrep/releases/tag/flowrep-0.5.2) is available and bump to use it.

liamhuber:
```
 error    libmamba Could not solve for environment specs
      ├─ flowrep =0.5.2 * is requested and can be installed;
      └─ semantikon =1.3.1 * is not installable because it requires
         └─ flowrep ==0.5.1 *, which conflicts with any installable versions previously reported.
```

liamhuber:
The next conda build of semantikon should have the upper bounds relaxed enough to accommodate the flowrep patch bump; it's already released, I'm just waiting for it to get on the github infrastructure.

Issue #865, opened 2026-06-18
`flowrep`-based `executorlib` interaction


liamhuber:
Naive in-session applications of `executorlib` should work straightforwardly -- we don't have any weird dynamic classes in the new implementation, so it should all be easier than ever for `executorlib` to digest. However, I need to dig into what's necessary to get persistence and run recovery across session shutdowns, and what that should even look like now that we've broken apart the workflow state from the actual run input/result data. Something akin to the legacy slurm tests should run.

liamhuber:
Closed by #875

PR #875, opened 2026-06-25
Executorlib QoL

liamhuber:
Extends the notebook with an executorlib section. As expected, naive usage works out-of-the-box thanks to stdlib `Executor`-interface compliance. 

For trans-shutdown re-discoverability of cached results, I followed a similar pattern as in the legacy code: a very thin wrapper around the `executorlib` executors which is expected to _only_ work with the `.run` paradigm here, and force-overloads the `cache_dir` and `cache_key` resource properties. In this case, `RunConfig.run_dir` is used to set the dir, determining uniqueness for rediscoverability, and the `node.lexical_path` is used for the individual result. That means re-running the same workflow with _different_ inputs at the same `run_dir` will get false cache hits. That's not my favourite, but at least this is simple and easy to understand: caches are made in the `run_dir` -- want a fresh cache location? Specify a new directory for your run. It avoids the complexity of trying to get hashes based on the input that are both totally unique _and_ which manage to get cache hits when you re-instantiate and re-run a workflow (e.g. with new instances of non-trivial input data, which wouldn't by default get hashed the same). I think this is a fair tradeoff, but we can always look to add more hand-holding later.

TODO: getting a version of the slurm test CI working for these. Should be a bit simpler and more robust because now we can add progress hooks to hard-crash at certain event points instead of having to rely on sleep times.

### DIFF
--- a/.github/workflows/push-pull.yml
+++ b/.github/workflows/push-pull.yml
@@ -38,3 +38,15 @@ jobs:
     uses: ./.github/workflows/slurm-test.yml
     with:
       mode: discover
+
+  slurm-interruption-wfms:
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/slurm-test-wfms.yml
+    with:
+      mode: interrupt
+
+  slurm-discovery-wfms:
+    if: github.event_name == 'pull_request'
+    uses: ./.github/workflows/slurm-test-wfms.yml
+    with:
+      mode: discover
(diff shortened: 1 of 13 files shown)


liamhuber:
The coverage complaint is for lines inside the overridden `.submit` method on the wrapped executors -- this absolutely _is_ running in the integration test, it's just that coverage is not picking up on it. I dug into it a bit, and couldn't solve it simply and didn't like the complexity of Claude's suggestion, so I'm just going to leave it uncovered. The critical thing is that if it stops working the test will fail, the gold star for the coverage report is just a nice-to-have. Still, I won't `pragma` it because it ought to be soluble, so missing the gold star is motivation to get to it one day.

liamhuber:
Closes #865

PR #876, opened 2026-06-25
New SLURM CI debugging

liamhuber:
I think it will probably take me a couple attacks to get the SLURM test for the new implementation all ironed out, and I don't want to clutter the main branch history; debug here, then take just what works back to #875


liamhuber:
Jobs are failing to ever appear in the queue in the new tests... After reactivating the legacy tests and inserting the new queue-printing method, there's nothing wrong with finding it this way. So either there is a real functional problem that they're not getting submit, or I'm just -- ah, yeah, the callbacks are run synchronously, so launching the termination callback is blocking us from proceeding to actually start the run. I'll dig into making the callbacks (optionally?) asynchronous

liamhuber:
@jan-janssen -- quick shout-out that the `executorlib` docs had exactly what I was looking for and it's appreciated. I got a uselessly opaque error message and finally I was like "well, let's unpack it, I wonder if Jan has any instructions on how to dig out more useful information" and that's [exactly what I found](https://executorlib.readthedocs.io/en/latest/2-hpc-cluster.html#slurmclusterexecutor).

Turned out my IDE had formatted the sbatch submission template with bad indentation. Was super easy to track down after my local equivalent of `print(f1.exception().output)` 🙏

liamhuber:
Incorporated into #875

PR #883, opened 2026-07-07
Add caching via Fleche

### LINKED ISSUE #866: `flowrep`-based caching
I removed caching altogether for now. Some sort of solution needs to be re-introduced.

liamhuber:
[`fleche`](https://github.com/pmrv/fleche/)-decorated functions are compatible with `flowrep`-/`pyiron_workflow`-decorated functions. For in-process runs (i.e. local execution and thread executors), fleche's memory cache will already get you speedup; for all cases (i.e. local and out-of-process executors), fleche-caching is triggered by passing a cache object to the run config:

```python
import pathlib
import shutil
import time

import fleche
from fleche.caches import Cache
from fleche.storage import ValuePickleFile, CallPickleFile

import pyiron_workflow._wfms.api as pwf

cache_dir = pathlib.Path(".pickle_cache")
pickle_cache = Cache(
    values=ValuePickleFile.with_pickle(root=cache_dir / "values"),
    calls=CallPickleFile.with_pickle(root=cache_dir / "calls"),
)

@pwf.atomic("slept_for")
@fleche.fleche
def sleepy(t):
    time.sleep(t)
    return t

n = sleepy.pwf.node()
out1 = n.run(pwf.RunConfig(fleche_cache=pickle_cache), t=1)
out2 = n.run(pwf.RunConfig(fleche_cache=pickle_cache), t=1)
print(out1.duration.total_seconds(), out2.duration.total_seconds())
>>> 1.011617 0.003676

shutil.rmtree(cache_dir, ignore_errors=True)
```

This differs from fleche's typical `with cache, executor:` syntax. Some changes were necessary so that we could handle executor _instructions_ (which are in turn necessary to nest executors). Simple local stuff comes for free, and you can still get the `with`-syntax working manually for un-nested executors, but I opted to maintain and present a single "we promise this will work" interface in the form of the run config field.

Fleche-decorating macros has no effect, since they don't actually invoke their decorated function when they run as a graph (they'll still be fleche-decorated functions if you invoke them as functions, of course). In principle very large graphs or very length flow-control loops might ostensibly run into seconds or minutes of runtime, but I think caching atomic nodes already covers the functionality that regularly risks running for minutes/hours/days.

Closes #866


liamhuber:
@pmrv, you're welcome to dig into the technicals as deeply as you'd like, but I pinged you because I'd like your opinion on activating fleche via providing a `Cache` to the run config.

pyiron_workflow/_wfms/execution.py
```diff
@@ -12,9 +13,16 @@
 from typing import TYPE_CHECKING, Any, Generic, NamedTuple, TypeAlias, TypeVar
 
 import flowrep as fr
+from pyiron_snippets import import_alarm
 
 from pyiron_workflow._wfms import lexical
 
+with import_alarm.ImportAlarm(
+    "Using a flech-cache requires 'fleche'.", raise_exception=True
```
pmrv:
```suggestion
    "Using a fleche-cache requires 'fleche'.", raise_exception=True
```

pmrv (commented):
I think as it is passing passing Cache() around and activating when required is fine, though I would have naively expected that just

```python
with fleche.cache(pickle_cache):
    out1 = n.run(t=1)
    out2 = n.run(t=1)
```

would also work, provided you are still wrapping the executors with `wrap_executor` when detecting fleche availability.

The only downside I see with passing it this way, is that if a user doesn't do it, but has fleche config files around, the fleche wrapped nodes would sometimes cache (when run in the notebook) and maybe under some executors, but not reliably.  On hand that's a bit of a footgun, on the other hand `RunConfig` is also a clearly defined point to opt-in.  If the keeping it, it may be cleaner to explicitly disable caching in this case.  There's `fleche.cache('void')` to do it, but also a bit annoying to always have to do it on the off-chance that some nodes may be fleche wrapped.  Your design call. :P


(I'm also taking away from this that the setup dance could be a bit cleaner on fleche's side.)

liamhuber:
> I think as it is passing passing Cache() around and activating when required is fine, though I would have naively expected that just
> 
> ```python
> with fleche.cache(pickle_cache):
>     out1 = n.run(t=1)
>     out2 = n.run(t=1)
> ```
> 
> would also work, provided you are still wrapping the executors with `wrap_executor` when detecting fleche availability.

Yes, this is correct, it seems to work fine that way.

> The only downside I see with passing it this way, is that if a user doesn't do it, but has fleche config files around, the fleche wrapped nodes would sometimes cache (when run in the notebook) and maybe under some executors, but not reliably. On hand that's a bit of a footgun, on the other hand `RunConfig` is also a clearly defined point to opt-in. If the keeping it, it may be cleaner to explicitly disable caching in this case. There's `fleche.cache('void')` to do it, but also a bit annoying to always have to do it on the off-chance that some nodes may be fleche wrapped. Your design call. :P

Hmm, `fleche.cache('void')` is interesting. What I wanted to supply was a guaranteed path to get it to work, I hadn't thought whether it would be annoying if you got accidental (but correct) cache hits. For now I think I'll leave it, but good to know the out is there.

> (I'm also taking away from this that the setup dance could be a bit cleaner on fleche's side.)

👍 Honestly, other than the always-bind-in-wrapped-executors to get nested fleche'd functions hitting, it has been a breeze to work with. My only real complaint is that I constantly get a warning message to the effect of "can't find a config file, falling back to in-memory cache". I'd rather this were just an info that most interpreters ignore than the warning (or whatever it is) that is so aggressively printed.

pmrv:
> > I think as it is passing passing Cache() around and activating when required is fine, though I would have naively expected that just
> > ```python
> > with fleche.cache(pickle_cache):
> >     out1 = n.run(t=1)
> >     out2 = n.run(t=1)
> > ```
> > 
> > would also work, provided you are still wrapping the executors with `wrap_executor` when detecting fleche availability.
> 
> Yes, this is correct, it seems to work fine that way.

For basic usage that's would make 'simplest' for users, imo.  That's also how I do it now mostly when using fleche.  I either set a config file or set caches up at the beginning of a notebook and then the rest of the code doesn't even see fleche.  I can still see the utility of RunConfig, especially if one has a few workflows or storage targets.

> > The only downside I see with passing it this way, is that if a user doesn't do it, but has fleche config files around, the fleche wrapped nodes would sometimes cache (when run in the notebook) and maybe under some executors, but not reliably. On hand that's a bit of a footgun, on the other hand `RunConfig` is also a clearly defined point to opt-in. If the keeping it, it may be cleaner to explicitly disable caching in this case. There's `fleche.cache('void')` to do it, but also a bit annoying to always have to do it on the off-chance that some nodes may be fleche wrapped. Your design call. :P
> 
> Hmm, `fleche.cache('void')` is interesting. What I wanted to supply was a guaranteed path to get it to work, I hadn't thought whether it would be annoying if you got accidental (but correct) cache hits. For now I think I'll leave it, but good to know the out is there.

👍

> > (I'm also taking away from this that the setup dance could be a bit cleaner on fleche's side.)
> 
> 👍 Honestly, other than the always-bind-in-wrapped-executors to get nested fleche'd functions hitting, it has been a breeze to work with. My only real complaint is that I constantly get a warning message to the effect of "can't find a config file, falling back to in-memory cache". I'd rather this were just an info that most interpreters ignore than the warning (or whatever it is) that is so aggressively printed.

Agree, fixed in the next release. :'D

liamhuber:
> For basic usage that's would make 'simplest' for users, imo. That's also how I do it now mostly when using fleche. I either set a config file or set caches up at the beginning of a notebook and then the rest of the code doesn't even see fleche. I can still see the utility of RunConfig, especially if one has a few workflows or storage targets.

So after digesting it, I actually swung back towards wanting to be stricter with the interface. Your suggestion actually did the trick in a single line modification:

```python
    def RunConfig._fleche_cache_context(self) -> BaseCache | contextlib.AbstractContextManager:
        return (
            fleche.cache(self.fleche_cache)
            if self.fleche_cache
            else fleche.cache("void")  # contextlib.nullcontext()
        )
```

This means that fleche-decorated functions work exactly as normal when you're using them as functions, but if you `.run` them with pwf, you only get caching if you explicitly set the cache in the run config.

It's honestly a foot-gun no matter what -- the `with` syntax can't work with executor instructions (instead of instances -- we need instructions to facilitate deep nesting), while the new more restrictive way means users naively expecting `with...` to work will be surprised. I came down on the side of being verbosely predictable.

pmrv:
Yeah, it's one spot to opt-in, so I think that's fair.

PR #886, opened 2026-07-22
Reformat `Run.outputs`

liamhuber:
This property was just a shortcut to `Run.result.output_ports`. First, it was unwise of me to make an alias to something but with a different name (`outputs` vs `output_ports`); second, I find this is just not actually the shortcut I want. If I want the whole ports map, I can just type the extra `.result.`. What I find I _do_ want very often is just quick dot/key-access to the result port _values_.

So now the `Run.outputs` property just builds a dot-dict out of the underlying port map `.value`s.


liamhuber:
Nevermind, rebasing on top of this is a huge pain, but the change here is very simple. I will wait for this until updating on top of https://github.com/pyiron/flowrep/pull/277 is finished

PR #887, opened 2026-07-22
Bump flowrep

liamhuber:
Includes reworking the dataclass decorator on top of flowrep's attack, moving the standard library of nodes to flowrep, and handling constants.


liamhuber:
Integration tests on dags using process pool executor instructions broke:

```
RuntimeError: os.fork is unsafe while filelock is changing descriptor ownership
```

It turns out that linux uses a different, faster but more dangerous startup routine than windows or macos. In `filelock-3.30.0` this danger got promoted to an explicit error. So, we've been safe on windows and macos all along, but previously had the ability to construct a dangerous executor case on linux.

I got Claude to patch it by modifying adding an extra flag to `ExecutorInstructions` and modifying its `.instantiate` to massage linux to behave like windows and macos.

From claude:

  ▎ Why CI broke and this didn't
  ▎
  ▎ pyiron_workflow evaluates sibling nodes in a DAG layer on separate threads (dag._multithreaded_layers). When those nodes carry their own ExecutorInstructions, each thread builds its own ProcessPoolExecutor — 
  ▎ so on Linux, where fork was the default start method, we were calling os.fork() concurrently from several threads of an already multi-threaded process.
  ▎
  ▎ That has always been unsafe: the child inherits a memory snapshot including locks held by threads that don't exist in it. CPython has warned about it for years (DeprecationWarning: This process is 
  ▎ multi-threaded, use of fork() may lead to deadlocks in the child), and CPython 3.14 changes the Linux default to forkserver for exactly this reason.
  ▎
  ▎ filelock 3.30.0 got there first. On import — no FileLock need ever be constructed — it installs a process-wide sys.addaudithook that raises RuntimeError: os.fork is unsafe while filelock is changing descriptor
  ▎ ownership if one thread forks while another fork is in flight. We import filelock transitively via fleche, which execution.py imports unconditionally. fleche pins filelock>=3.4, and the CI conda cache key 
  ▎ includes the date, so a routine daily re-solve picked up ≥3.30 and turned a latent hazard into two hard failures. Nothing in this repo changed.
  ▎
  ▎ It never reproduced on macOS or Windows because both default to spawn and never call os.fork at all. Linux was the only platform on the fast-but-dangerous path.
  ▎
  ▎ The fix
  ▎
  ▎ ExecutorInstructions.instantiate() is the one place the library owns the construction of a user's process pool, so that's where the fix goes: when the constructor is a ProcessPoolExecutor subclass and the 
  ▎ caller hasn't supplied their own mp_context, we inject mp_context=get_context("spawn") — putting Linux on the same footing as macOS and Windows instead of leaving it uniquely exposed. An explicit mp_context 
  ▎ always wins, and start_method=None opts out.
  ▎
  ▎ This addresses the actual failure mode rather than the symptom. Two alternatives were tried and rejected:
  ▎
  ▎ - Pinning filelock<3.30 — hides a real bug and fights the ecosystem.
  ▎ - Serialising the forks with a lock in _multithreaded_layers — measurably worse. Baseline was 15/15 clean; with the lock the suite hung on run 4 of 5, and again with os.register_at_fork(after_in_child=...) 
  ▎ re-initialisation. Re-initialising our own lock doesn't help, because it was never the only one the child inherited. Serialising silences the error and keeps the deadlock.
  ▎
  ▎ Live user-supplied executor instances were never at risk: CPython holds self._shutdown_lock across _adjust_process_count() in ProcessPoolExecutor.submit, so a single pool cannot fork concurrently with itself. 
  ▎ The failure requires distinct pool objects, which is precisely the ExecutorInstructions path.
  ▎
  ▎ Cost: non-fork workers re-import the module tree (~1.25s for the pyiron_workflow stack), so the overhead is roughly (nested process generations) × import time. The integration suite goes 39.9s → 45.8s. If that
  ▎ becomes a problem, multiprocessing.set_forkserver_preload([...]) with start_method="forkserver" recovers a large chunk of it (measured: 3.50s → 2.43s on the worst single test).
  ▎
  ▎ Note for local runs: spawn workers don't resolve relative PYTHONPATH entries the way fork children did. CI is unaffected (add-to-python-path uses absolute paths), but a hand-rolled relative PYTHONPATH on Linux
  ▎ will now fail with ModuleNotFoundError.

liamhuber:
```
  FAIL: test_local_pool_instructions (integration._wfms.test_caching.TestFlecheCaching.test_local_pool_instructions)
    File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/integration/_wfms/test_caching.py", line 112, in test_local_pool_instructions
      self._assert_speedup(self._no_executors, self._executor_instructions)
    File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/integration/_wfms/test_caching.py", line 63, in _assert_speedup
      self.assertLess(
  AssertionError: 6.112269 not less than 2.5 : Warm run took 6.11s; expected < 2.5s sleep (cache hit should skip the sleep)
  
```

Ok, this is because the test is conflating the pool spinup time + sleep time with the sleep time. If the pool spinup is larger than the sleeping time, we'll get sunk here regardless of whether we got a cache hit (since cache hits don't currently short-circuit starting up the executor). I.e., I believe this is just a weakness in the test logic.

PR #903, opened 2026-07-27
Purge old version.

liamhuber:
The readme hard-links to the GitHub page for the last version with compatiblity features.

This trims a number of packages off the dependencies.


liamhuber:
I just checked on codecov, and this brings the tracked lines in `pyiron_workflow` down to 2300 from 6100, which I'm happy about.

PR #911, opened 2026-07-28
Bump typeguard and flowrep

liamhuber:


liamhuber:
```

      ├─ semantikon =1.4.0 * is installable and it requires
      │  └─ typeguard >=4.0.0,<4.6.0 *, which can be installed;
      └─ typeguard =4.6.0 * is not installable because it conflicts with any installable versions previously reported.

```

not a big deal, the upgrade is not in any way urgent.

PR #922, opened 2026-08-12
Extend support to 3.14

liamhuber:


liamhuber:
```
  FAIL: test_partial_state_preserved_in_for_each (integration.test_failure.TestForEachFailure.test_partial_state_preserved_in_for_each)
    File "/home/runner/work/pyiron_workflow/pyiron_workflow/tests/integration/test_failure.py", line 265, in test_partial_state_preserved_in_for_each
      self.assertEqual(for_each_step.steps[1].outputs.m, 0)
      ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  AssertionError: 1 != 0

```

Ah, indeed. In the test we assume that step order and body node index order are the same. But when we're running with multithreaded dag layers (which in this case we are), the different body nodes might race each other to append to the `steps` field.

The `steps` are anyhow just a shortcut to "what got run in order", and inside a DAG layer there is no principle to intrinsically demand a particular temporal order -- that's the point of they layers, they are execution-order-independent, so there's no reason to insist on a particular order they get appended to `steps`. My intuition is to simply say the test is structured poorly and is trying to test the wrong thing. 

I verified that manually disabling DAG layer multithreading resolves the issue. If you want to demand that your steps follow the same order as iterating through each layer, then that's my suggested solution. Here, I'll just fix the test.

The only outstanding thing -- why didn't it ever hit this race in python 3.12? Because this test has been run hundreds of times on that version and never failed, so even if was _in principle_ stochastic, in practice it was completely stable. So what happened in the intervening python versions?

liamhuber:
I asked Claude to look into it and find the answer satisfactory

>  It isn't ThreadPoolExecutor semantics. I diffed concurrent/futures/thread.py 3.13→3.14: the only change is the WorkerContext refactor that generalises the class for PEP 734's InterpreterPoolExecutor. Scheduling,
  the work queue, and the _idle_semaphore reuse logic are byte-for-byte equivalent.
>
>  The trigger is structural, in your code: _multithreaded_layers uses one executor for all layers. So when the body layer is submitted, a worker from the scatter layer is already alive and parked in
  work_queue.get(block=True). That means:
>
>  - submit(body_0) → _adjust_thread_count acquires _idle_semaphore, spawns nothing, and hands the item to the parked thread, which must be woken through the queue's condvar and then take the GIL off the
  still-submitting main thread.
>  - submit(body_1..5) → each spawns a fresh thread that starts running straight away.
>
>  So body_0 is the one node in the layer running on a recycled thread, racing the main thread's submit loop. That's the whole race.

samwaseda:
finally!!

PR #927, opened 2026-08-21
Don't pin python in the env yamls

liamhuber:
Mybinder is hung up on it, and the pyproject.toml and conda feedstock manage pinning python versions for us.


liamhuber:
It still needs to appear in the binder env somewhere, so I stuck a python pin in the notebooks env file (which the auto-update uses to populate the binder env and doesn't hurt). Confirmed on the branch mybinder that this resolves the problem.

I wonder if this is a recent mybinder change, or if I just haven't checked since specifying a python range? Either way, I guess the problem exists now on other repos.

